PR #4, opened 2021-06-09
Fix item and vault loading by name

florisvdg:
More bash 😅, and also am aware that the amount of API calls could be reduced.
But a large chunk of this script should be removed soon anyway.

Also added a test for it.

Took the UUID determination logic from [here](https://github.com/1Password/onepassword-operator/blob/b574e394ad649d2cbf6f61192f00a1918cef8df3/pkg/onepassword/uuid.go#L7).


jpcoenen (approved):
Nice, that's a lot easier to use!

PR #5, opened 2021-08-03
Add the op CLI in the script

edif2008:
By adding the 1Password Command Tool in the script, we no longer need to look for environment variables that have a reference and fetch the values of the secrets through shell script. Instead, we use the commands `op list envars` and `op read` (alpha version)

Side note: this is dependent on the release of the version of the 1Password Command Line Tool that contains the command `op list envars`. Once that is released, this can be merged.


entrypoint.sh
```diff
@@ -2,6 +2,10 @@
 # shellcheck disable=SC2046,SC2001,SC2086
 set -e
 
+# Install op-cli
+curl -sSfLo op.zip "https://drive.google.com/uc?export=download&id=1HRAsihTN0Cx0pWZEWN06jAWxo0eW5eG-"
```
jillianwilson, line 6: I wonder if we should instead pull in a dockerfile with the cli already installed, rather than downloading and installing the cli each time

edif2008: That may sound like a better approach. If we go this way, then we should have a tag for the alpha release for the `1password/op` image (which has the commands we need)

entrypoint.sh
```diff
@@ -27,111 +31,32 @@ if [ "$INPUT_UNSET_PREVIOUS" == "true" ]; then
...
+  # Register a mask for the secret to prevent accidental log exposure.
+  # To support multiline secrets, escape percent signs and add a mask per line.
+  escaped_mask_value=$(echo "$secret_value" | sed -e 's/%/%25/g')
+  IFS=$'\n'
+  for line in $escaped_mask_value; do
+    if [ "${#line}" -lt 3 ]; then
+      # To avoid false positives and unreadable logs, omit mask for lines that are too short.
+      continue
+    fi
+    echo "::add-mask::$line"
```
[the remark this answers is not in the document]
edif2008: My testing environment was wrong. [Act](https://github.com/nektos/act) does not support masking, thus the confusion whether my implementation made the masking properly or not.

Now the next step is the following: identify whether a certain field is concealed or not.

entrypoint.sh
SimonBarendse: We'll push to https://hub.docker.com/r/1password/op once the functionality hits GA.

We also provide a [Dockerfile](https://bucket.agilebits.com/cli-private-beta/v2/Dockerfile) that you can already use during the alpha period.

jillianwilson (approved):

entrypoint.sh
edif2008: In the current approach (a Javascript action), it is not possible to use a dockerfile for the following reasons:
- Lose compatibility with MacOS runners (and Windows runners if we will make the action compatible with it in the future)
- `GITHUB_ENV` does not exist in docker container, thus the user's secrets cannot be loaded.

The only maintenance we have to do in regards to the CLI is the link from where we download it, which can easily be parameterized if the link structure is similar between different versions.

edif2008:
Putting this PR on hold for now, since it is waiting for the functionality of the CLI to hit GA.

README.md
```diff
@@ -1,11 +1,90 @@
...
+      - name: Load secret
+        id: op-load-secret
+        uses: 1password/load-secrets-action@v1
+        env:
+          OP_CONNECT_HOST: <Your Connect instance URL>
+          OP_CONNECT_TOKEN: ${{ secrets.OP_CONNECT_TOKEN }}
+          SECRET: op://app-cicd/hello-world/secret
+
+      - name: Print masked secret
+        run: echo "Secret: ${{ steps.op-load-secret.outputs.SECRET }}"
```
jillianwilson, line 37: Can we use a real world example here, rather than echo?

jillianwilson (approved):

edif2008:
**Note:** Find a fix for command injection, specifically [this code](https://github.com/1Password/load-secrets-action/pull/5/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R14).

hanerd:
Is there anything I can do to help get this merged in?

We make extensive use of `op inject ` and would love to put up a PR to include that in this action.

entrypoint.sh
```diff
@@ -2,153 +2,133 @@
...
-    echo "Expected path to be in format op://<vault>/<item>[/<section>]/<field>: $ref"
-    continue
+}
+
+# Install op-cli
+install_op_cli() {
+  if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    curl -sSfLo op.zip "https://cache.agilebits.com/dist/1P/op2/pkg/v2.6.0-beta.06/op_linux_amd64_v2.6.0-beta.06.zip"
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    curl -sSfLo op.zip "https://cache.agilebits.com/dist/1P/op2/pkg/v2.6.0-beta.06/1password-cli_v2.6.0-beta.06_darwin_amd64.tar.gz"
```
edif2008, line 36: Does it really work this way? Downloading a `.tar.gz` and save it as a `.zip`...I assume a different type of archive works differently, and changing the file extension may lead to some unexpected surprises.

entrypoint.sh
```diff
@@ -2,153 +2,133 @@
...
   fi
-
-  path=$(echo $ref | sed -e "s/^op:\/\///")
-  if [ $(echo "$path" | tr -cd '/' | wc -c) -lt 2 ]; then
-    echo "Expected path to be in format op://<vault>/<item>[/<section>]/<field>: $ref"
-    continue
+}
+
+# Install op-cli
+install_op_cli() {
```
edif2008, line 32: In the future it would be nice to support Windows runners as well

action.yml
```diff
@@ -8,11 +8,9 @@ inputs:
...
+    default: false
 runs:
-  using: composite
-  steps:
-    - shell: bash
-      env:
-        INPUT_UNSET_PREVIOUS: ${{ inputs.unset-previous }}
-      run: |
-        ${{ github.action_path }}/entrypoint.sh
+  using: 'node12'
```
edif2008, line 15: Change to `node16`, since that is the most recent one.

README.md
```diff
@@ -1,11 +1,99 @@
...
+
 
 Specify right from your workflow YAML which secrets from 1Password should be loaded into your job, and the action will make them available as environment variables for the next steps.
 
+## Prerequisites 
+ - [1Password Connect](https://support.1password.com/secrets-automation/#step-2-deploy-a-1password-connect-server) deployed in your infrastructure
+
 ## Usage
 
+You can configure the action to use either 1Password connect instance or service account.
```
edif2008, line 13: Connect is capitalized, since it is the name of a product we've built.
```suggestion
You can configure the action to use either 1Password Connect instance or service account.
```

README.md
```diff
@@ -1,11 +1,99 @@
 # Load Secrets from 1Password - GitHub Action
 
-The action to load secrets from [1Password Connect](https://1password.com/secrets/) into GitHub Actions.
+This action loads secrets from [1Password Connect](https://1password.com/secrets/) into GitHub Actions.
```
edif2008, line 3: This is no longer true. We can load secrets directly from 1Password with service accounts.

Maybe this phrasing is better?
```suggestion
This action loads secrets from 1Password into GitHub Actions using [1Password Connect](https://1password.com/secrets/) or a Service Account.
```

README.md
```diff
@@ -1,11 +1,99 @@
...
+
 
 Specify right from your workflow YAML which secrets from 1Password should be loaded into your job, and the action will make them available as environment variables for the next steps.
 
+## Prerequisites 
+ - [1Password Connect](https://support.1password.com/secrets-automation/#step-2-deploy-a-1password-connect-server) deployed in your infrastructure
```
edif2008, line 9: This is no longer a prerequisite, since it only applies when the users want to use the action with Connect.

README.md
```diff
@@ -1,11 +1,99 @@
...
+
 ## Usage
 
+You can configure the action to use either 1Password connect instance or service account.
+
+If provide `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables - connect instance will be used to load secrets.
+
+If provide `OP_SERVICE_ACCOUNT_TOKEN` variable - service account will be used to load secrets.
```
edif2008, line 17:
```suggestion
If you provide `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables, the Connect instance will be used to load secrets.

If you provide `OP_SERVICE_ACCOUNT_TOKEN` variable, the service account will be used to load secrets.
```

README.md
```diff
@@ -1,11 +1,99 @@
...
+
+***Note***: if all variables are provided the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` environment variables will take precedence over `OP_SERVICE_ACCOUNT_TOKEN`. Clear the Connect environment variables to configure a service account instead.
```
edif2008, line 19: Maybe this phrasing highlights better the message?

```suggestion
***Note***: if all environment variables are provided, the GitHub action will use Connect over the service account. Clear the Connect environment variables to make the action use a service account instead.
```

README.md
```diff
@@ -107,16 +204,14 @@ So for example, the reference URI `op://app-cicd/aws/secret-access-key` would be
...
-This means that a username or port field for example will not get masked.
-
-## 1Password Connect Configuration
+## 1Password Configuration
 
-To use the action, you need to have a [1Password Connect](https://support.1password.com/secrets-automation/#step-1-set-up-a-secrets-automation-workflow) instance deployed somewhere.
-To configure the action with your Connect URL and a Connect token, you can set the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables.
+To use the action, you may need to have a [1Password Connect](https://support.1password.com/secrets-automation/#step-1-set-up-a-secrets-automation-workflow) instance deployed somewhere.
+To configure the action with your Connect URL and a Connect token, you can set the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables. 
+If you provide `OP_SERVICE_ACCOUNT_TOKEN` you can skip connect setup as action will use your service account instead.
```
edif2008, line 214: This part is a bit confusing. Does the user really need 1Password Connect deployed or not? Maybe this suggestion says it clearer when it's needed? Also, I've rephrased some of it to be consistent.

```suggestion
To use the action with Connect, you need to have a [1Password Connect](https://support.1password.com/secrets-automation/#step-1-set-up-a-secrets-automation-workflow) instance deployed somewhere.
To configure the action with your Connect host and token, set the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` environment variables. 
To configure the action with your service account token, set the `OP_SERVICE_ACCOUNT_TOKEN` environment variable.
```

entrypoint.sh
```diff
@@ -2,153 +2,133 @@
...
+  IFS=$'\n'
+  for env_var in $(op env ls); do
+    populating_secret $env_var
+  done
+}
 
-  jq_secret_selector="$jq_section_selector and ($jq_field_selector)"
+# Load environment variables using connect service. Iterate over hem to find 1Password references, load the secret values,
+# and make them available as environment variables in the next steps.
+extract_using_connect() {
```
edif2008, line 100: From this part it gives me the impression that `op env ls` doesn't work with Connect. This shouldn't be the case. I've raised this bug and hopefully the upcoming beta releases will fix this 😄 
In this way, this code will be simplified quite a bit.

.github/workflows/test.yml
```diff
@@ -15,18 +15,70 @@ jobs:
...
+        with:
+          export-env: true
         env:
-          MULTILINE_SECRET: op://v5pz6venw4roosmkzdq2nhpv6u/ghtz3jvcc6dqmzc53d3r3eskge/notesPlain
+          SECRET: op://acceptance-tests/test-secret/password
+          SECRET_IN_SECTION: op://acceptance-tests/test-secret/test-section/password
+          MULTILINE_SECRET: op://acceptance-tests/multiline-secret/notesPlain
+      - name: Assert test secret values
+        run: ./tests/assert-env-set.sh
+  test-connect-export:
```
edif2008, line 59: How is this job different from `use-connect-with-export-env`? From the looks of it, the only difference is that this job also has the unsetting step. Maybe they could be merged into one. Let me know how you feel about this.

edif2008 (commented):
Code review: ✅ 
Looking strong. My comments are a bunch of nitpicks in documentation, as well as some code to make sure it's all good.

entrypoint.sh
volodymyrZotov: changed the way to correctly extract `.tar.gz`

entrypoint.sh
volodymyrZotov: Agree

action.yml
volodymyrZotov: changed

README.md
volodymyrZotov: fixed

README.md
volodymyrZotov: Agree. Changed.

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

.github/workflows/test.yml
volodymyrZotov: yeah, makes sense to me

.github/workflows/test.yml
```diff
@@ -15,37 +15,84 @@ jobs:
...
-          connect-host: http://localhost:8080
+          connect-host:	http://localhost:8080
           connect-token: ${{ secrets.OP_CONNECT_TOKEN }}
       - name: Load secrets
+        id: load_secrets
         uses: ./ # 1password/load-secrets-action@<version>
         env:
-          SECRET: op://v5pz6venw4roosmkzdq2nhpv6u/hrgkzhrlvscomepxlgafb2m3ca/password
-          SECRET_IN_SECTION: op://v5pz6venw4roosmkzdq2nhpv6u/hrgkzhrlvscomepxlgafb2m3ca/test-section/password
-          UNMASKED_VALUE: op://v5pz6venw4roosmkzdq2nhpv6u/hrgkzhrlvscomepxlgafb2m3ca/test-section/username
```
edif2008, line 25: I see that the tests where the references have IDs have been removed. Any reason for doing so?
As a dev, this would give me confidence that both options work well with the action.

Let me know if there's anything you need to bring back these tests (if you feel like it makes sense to include them).

volodymyrZotov: I tried to use with those ids, but test fails with the error ```status 403: Authorization: token does not have access to vault v5pz6venw4roosmkzdq2nhpv6u```

README.md
```diff
@@ -1,11 +1,28 @@
...
 
-The action to load secrets from [1Password Connect](https://1password.com/secrets/) into GitHub Actions.
+This action loads secrets from 1Password into GitHub Actions using [1Password Connect](https://1password.com/secrets/) or a Service Account.
+
 
 Specify right from your workflow YAML which secrets from 1Password should be loaded into your job, and the action will make them available as environment variables for the next steps.
 
 ## Usage
 
+You can configure the action to use either 1Password Connect instance or service account.
```
jillianwilson, line 10: Would change this to: ```You can configure the action to use either 1Password Connect instance or a 1Password Service Account. Service Accounts are currently in Beta and are only available to select users.```

README.md
```diff
@@ -1,11 +1,28 @@
...
+
+If you provide `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables, the Connect instance will be used to load secrets. Make sure [1Password Connect](https://support.1password.com/secrets-automation/#step-2-deploy-a-1password-connect-server) deployed in your infrastructure.
+
+If you provide `OP_SERVICE_ACCOUNT_TOKEN` variable, the service account will be used to load secrets.
+
+***Note***: if all environment variables are provided, the GitHub action will use Connect over the service account. Clear the Connect environment variables to make the action use a service account instead.
```
jillianwilson, line 16: I recommend changing this line to: ```If all environment variables have been set, the Connect credentials will take precedence over the provided service account token. You must unset the Connect environment variables to ensure the action uses the service account token.```

README.md
```diff
@@ -1,11 +1,28 @@
...
+
+There are two ways that secrets can be loaded:
+ - [use the secrets from the action's ouput](#use-secrets-from-the-actions-output)
+ - [export secrets as environment variables](#export-secrets-as-environment-variables)
+
+### Use secrets from the action's output
+
+This approach enables the user to use the loaded secrets as an output from the step: `steps.step-id.outputs.secret-name`. You need to set an id for the step that uses this action to be able to access its outputs. More details about the metadata syntax [here](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputsoutput_id).
```
jillianwilson, line 24: Change to ```This method allows for you to use the loaded secrets as an output from the step: `steps.step-id.outputs.secret-name`. You will need to set an id for the step that uses this action to be able to access its outputs. More details about the metadata syntax [here](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputsoutput_id).```

README.md
```diff
@@ -15,12 +32,83 @@ jobs:
...
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: acme/app:latest
+```
+</details>
+
+### Export secrets as environment variables
+
+In this approach, the user can access the loaded secrets as environment variables. These environment variables are accessible at a job level.
```
jillianwilson, line 92: Change to: ```This method, allows the action to access the loaded secrets as environment variables. These environment variables are accessible at a job level.```

README.md
```diff
@@ -107,16 +201,14 @@ So for example, the reference URI `op://app-cicd/aws/secret-access-key` would be
...
-This means that a username or port field for example will not get masked.
-
-## 1Password Connect Configuration
+## 1Password Configuration
 
-To use the action, you need to have a [1Password Connect](https://support.1password.com/secrets-automation/#step-1-set-up-a-secrets-automation-workflow) instance deployed somewhere.
-To configure the action with your Connect URL and a Connect token, you can set the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables.
+To use the action with Connect, you need to have a [1Password Connect](https://support.1password.com/secrets-automation/#step-1-set-up-a-secrets-automation-workflow) instance deployed somewhere.
+To configure the action with your Connect host and token, set the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` environment variables.  
+To configure the action with your service account token, set the `OP_SERVICE_ACCOUNT_TOKEN` environment variable.
```
jillianwilson, line 211: WE should add a note here again that service accounts are currently in beta and are only available to select users at this time

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: changed

README.md
volodymyrZotov: added a note

jillianwilson (approved):

.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
+.vscode
+.idea
+.secrets
```
edif2008, line 4: I don't think that we put IDE-related directories in our `.gitignore` files in other repositories.

To fix this on your local machine, you can make a global `.gitignore` and add those there. Then Git will never complain about them in any repo. 😄

README.md
```diff
@@ -1,11 +1,28 @@
...
 
-The action to load secrets from [1Password Connect](https://1password.com/secrets/) into GitHub Actions.
+This action loads secrets from 1Password into GitHub Actions using [1Password Connect](https://1password.com/secrets/) or a Service Account.
+
 
 Specify right from your workflow YAML which secrets from 1Password should be loaded into your job, and the action will make them available as environment variables for the next steps.
 
 ## Usage
 
+You can configure the action to use either 1Password Connect instance or a 1Password Service Account. Service Accounts are currently in Beta and are only available to select users.
```
edif2008, line 10: We should mention that Service Accounts are in beta. Something like this could do it:
```suggestion
You can configure the action to use either 1Password Connect instance or a 1Password Service Account (currently in beta). Service Accounts are currently in Beta and are only available to select users.
```

We can also bold it if needed.

README.md
```diff
@@ -133,7 +226,7 @@ jobs:
         with:
           connect-host: <Your Connect instance URL>
           connect-token: ${{ secrets.OP_CONNECT_TOKEN }}
-
+          service-account-token: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
```
edif2008, line 229: Adding both configurations in this example is not a great practice. I'd rather provide two examples: one with Connect and one with Service Accounts instead.

Another solution is to just remove it from this example.

README.md
```diff
@@ -1,11 +1,28 @@
...
+This action loads secrets from 1Password into GitHub Actions using [1Password Connect](https://1password.com/secrets/) or a Service Account.
+
 
 Specify right from your workflow YAML which secrets from 1Password should be loaded into your job, and the action will make them available as environment variables for the next steps.
 
 ## Usage
 
+You can configure the action to use either 1Password Connect instance or a 1Password Service Account. Service Accounts are currently in Beta and are only available to select users.
+
+If you provide `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables, the Connect instance will be used to load secrets. Make sure [1Password Connect](https://support.1password.com/secrets-automation/#step-2-deploy-a-1password-connect-server) deployed in your infrastructure.
```
edif2008, line 12:
```suggestion
If you provide `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables, the Connect instance will be used to load secrets. Make sure [1Password Connect](https://support.1password.com/secrets-automation/#step-2-deploy-a-1password-connect-server) is deployed in your infrastructure.
```

README.md
```diff
@@ -1,11 +1,28 @@
...
+
+If you provide `OP_SERVICE_ACCOUNT_TOKEN` variable, the service account will be used to load secrets.
```
edif2008, line 14: Add note about it being in beta, similar to this note used later in the readme:
_Service Accounts are currently in Beta and are only available to select users._

README.md
```diff
@@ -1,11 +1,28 @@
...
+
+***Note***: If all environment variables have been set, the Connect credentials will take precedence over the provided service account token. You must unset the Connect environment variables to ensure the action uses the service account token.
+
+There are two ways that secrets can be loaded:
+ - [use the secrets from the action's ouput](#use-secrets-from-the-actions-output)
+ - [export secrets as environment variables](#export-secrets-as-environment-variables)
+
+### Use secrets from the action's output
+
+This method allows for you to use the loaded secrets as an output from the step: `steps.step-id.outputs.secret-name`. You will need to set an id for the step that uses this action to be able to access its outputs. More details about the metadata syntax [here](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputsoutput_id).
```
edif2008, line 24: I'd rephrase a bit the end:
```suggestion
This method allows for you to use the loaded secrets as an output from the step: `steps.step-id.outputs.secret-name`. You will need to set an id for the step that uses this action to be able to access its outputs. For more details, see [`outputs.<output_id>` metadata syntax](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputsoutput_id).
```

edif2008 (commented):
I've taken another round through the code and now I feel more confident about this transition of the action.

One small remark I have (besides my comments):
I've noticed that we had a test when the action was used twice (e.g. load some secrets, then load some more). Now we no longer have it.
Would it help to have such test case? What do you think?

Once the comments have been addressed, this PR is good to be merged. 😄

README.md
volodymyrZotov: We mention this in the next sentence.
> Service Accounts are currently in Beta and are only available to select users.

.gitignore
volodymyrZotov: removed from repo's `.gitignore`

README.md
volodymyrZotov: removed

README.md
volodymyrZotov: fixed

README.md
volodymyrZotov: There is already such a note on line 9 above
> You can configure the action to use either 1Password Connect instance or a 1Password Service Account. Service Accounts are currently in Beta and are only available to select users.

README.md
volodymyrZotov: fixed

volodymyrZotov:
> 



> I've taken another round through the code and now I feel more confident about this transition of the action.
> 
> One small remark I have (besides my comments): I've noticed that we had a test when the action was used twice (e.g. load some secrets, then load some more). Now we no longer have it. Would it help to have such test case? What do you think?
> 
> Once the comments have been addressed, this PR is good to be merged. 😄

@edif2008 Thank you for the review.
Regarding the tests. Currently we have few more test cases and they cover what was initially and some new cases in addition.

PR #6, opened 2021-08-03
Enable the user to use the GitHub Action without Connect

edif2008:
In order to use this GitHub Action without Connect, we will make use of the `op signin` to sign the user in and fetch the desired secrets with `op read` (alpha version).
To do so, the user has to provide the following:
 - `user-domain` - the domain of the account
 - `user-email` - the user's account email
 - `user-key` - the user's secret key
 - `user-pwd` - the user's Master Password.

We also now have a new config variable, `use-connect`, which if set, will use the Connect instance set with the following variables: `connect-host`, `connect-token`.


edif2008:
Closed because it is made on a use case that will never be reached

PR #10, opened 2022-04-21
fix: Allow vaults and items with brackets

amontalban:
We are having issues when using this action because we have brackets in our vault and items names:

```
##[debug]/usr/bin/bash --noprofile --norc -e -o pipefail /home/runner/work/_temp/1afdd4f7-f7cf-41ca-a690-fbc13169b563.sh
Populating variable: SENTRY_AUTH_TOKEN
Getting vault ID from vault name: [STAGE] - Secrets
curl: (22) The requested URL returned error: 400 Bad Request
Could not find vault ID for vault: 
Error: Process completed with exit code 1.
```

So, this leverages on cURL feature to perform URL-encoding as described [here](https://stackoverflow.com/a/2027690).

I tested it with our own vaults and secrets, but it should cover other cases as well.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -74,7 +74,7 @@ for possible_ref in $(printenv | grep "=op://" | grep -v "^#"); do
 
   if [[ $(echo -n $(echo $vault | grep "^[a-z0-9]*$") | wc -c) -ne 26 ]]; then
     echo "Getting vault ID from vault name: $vault"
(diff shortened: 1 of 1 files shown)


edif2008:
Hey there. 
Thank you for helping us in making this GitHub Action better.

We're currently working on [this PR](https://github.com/1Password/load-secrets-action/pull/5) that enhances the way secrets are loaded. With that, we remove the `curl` commands that were used to fetch the secrets.

Would that approach address your issue as well?

edif2008:
Closing this PR for the following reasons:
- Brackets are considered illegal characters according to the [secret reference syntax rules](https://developer.1password.com/docs/cli/secrets-reference-syntax#syntax-rules).
- This can be mitigated by using item/vault IDs instead of names.

PR #17, opened 2022-10-03
Update readme to link to developer portal

libutcher:
This MR updates the readme to:
- Match the final edited text used in the developer portal
- Link out to the developer portal for usage examples beyond basic setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,66 @@ If you're using the action more than once in a single job, [you can use the `con
 
 You can load secrets using the action in two ways:
 
-1. [Use secrets from the action's output](https://developer.1password.com/docs/connect/github-actions/#use-secrets-from-the-actions-output)
-2. [Export secrets as environment variables](https://developer.1password.com/docs/connect/github-actions/#export-secrets-as-environment-variables)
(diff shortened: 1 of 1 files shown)


edif2008:
The README will also be used in the GitHub Marketplace, which is the place where folks may look for this action. Having the README page as suggested in this PR seems incomplete without some basic usage examples.

What I suggest is to provide the key snippets from the portal into this readme and then hint to the portal for more.
What do you think?

libutcher:
> The README will also be used in the GitHub Marketplace, which is the place where folks may look for this action. Having the README page as suggested in this PR seems incomplete without some basic usage examples.
> 
> What I suggest is to provide the key snippets from the portal into this readme and then hint to the portal for more. What do you think?

Sure, I'll add those in then ping you for a second review!

libutcher:
> The README will also be used in the GitHub Marketplace, which is the place where folks may look for this action. Having the README page as suggested in this PR seems incomplete without some basic usage examples.
> 
> What I suggest is to provide the key snippets from the portal into this readme and then hint to the portal for more. What do you think?

Done! Let me know if this looks better to you.

edif2008:
Closing this as we will use a new template on the README.

PR #21, opened 2022-12-12
Quote $ref to avoid word splitting

### LINKED ISSUE #20: Workflow fails if secret ref includes spaces
# Steps to reproduce

1. Create a vault called "Sample Vault"
2. Create an item called "SampleItem"
3. Give it a password value
4. Add this step in your Github Action:

```yml
 -  name: Load secrets
    uses: 1password/load-secrets-action@v1.1.0
    env:
      MY_SECRET: "op://Sample Vault/SampleItem/password"
```

You'll see the following error message when your action runs:

```
Populating variable: MY_SECRET
Error:  [timestamp] expected at most 1 arguments but got 2 instead
```

I believe the issue is that `$ref` needs quoting [here](https://github.com/1Password/load-secrets-action/blob/747c0b597462b5ffa8a6e332b8f4b068fcad8032/entrypoint.sh#L46). There are some other examples where quoting might be needed; I'd suggest removing (some of) the disabled shellcheck checks [here](https://github.com/1Password/load-secrets-action/blob/747c0b597462b5ffa8a6e332b8f4b068fcad8032/entrypoint.sh#LL2C4-L2C4).

---

I'm aware that we can work around this by using the Vault UUID in this instance, instead of "Sample Vault". But that solution is only available where the spaces are in the vault or item name. Section IDs appear to be the same as their names, even if there is a space in the name. e.g. for an item with an "add more" section:

```command
$ op item get --format json df4h6uiz5ownudtzdfse4ouse4 | jq '.sections'
[
  {
    "id": "add more"
  }
]
```

So if there is a space in the section name, this issue can only be resolved by quoting `$ref` when passing to `op read`.

simonwhitaker:
Fixes #20

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -43,7 +43,7 @@ populating_secret() {
   ref=$(printenv $1)
 
   echo "Populating variable: $1"
-  secret_value=$(op read $ref)
+  secret_value=$(op read "$ref")
 
   if [ -z "$secret_value" ]; then
     echo "Could not find or access secret $ref"


volodymyrZotov (approved):
Looks good! Thank you! 🚀

volodymyrZotov:
@simonwhitaker Could you please sign the commit? We can't merge it.

simonwhitaker:
> @simonwhitaker Could you please sign the commit? We can't merge it.

Done!

PR #25, opened 2022-12-21
Ensure that the action is backwards-compatible

edif2008:
This PR brings 2 changes that ensure that the GitHub Action is backwards compatible:
- Append `http://` if the prefix is not provided in the `OP_CONNECT_HOST` (this is caused by the fact that `curl` [guesses the protocol if not provided](https://linux.die.net/man/1/curl), which we missed when switching to using the 1Password CLI as the backend of the action)
- Set the default of `export-env` to `true`, since that was the default behavior of the action until we added the possibility to export secrets as step's output.

Also, the documentation is adjusted to reflect these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ jobs:
       - name: Load secret
         id: op-load-secret
         uses: 1password/load-secrets-action@v1
+        with:
+          export-env: false
         env:
           OP_CONNECT_HOST: <Your Connect instance URL>
           OP_CONNECT_TOKEN: ${{ secrets.OP_CONNECT_TOKEN }}
@@ -77,6 +79,8 @@ jobs:
       - name: Load Docker credentials
         id: load-docker-credentials
         uses: 1password/load-secrets-action@v1
+        with:
+          export-env: false
         env:
           OP_CONNECT_TOKEN: ${{ secrets.OP_CONNECT_TOKEN }}
           DOCKERHUB_USERNAME: op://app-cicd/docker/username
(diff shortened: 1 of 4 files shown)


entrypoint.sh
```diff
@@ -14,6 +14,10 @@ auth_type=$CONNECT
 managed_variables_var="OP_MANAGED_VARIABLES"
 IFS=','
 
+if [[ "$OP_CONNECT_HOST" != "http://"* ]] && [[ "$OP_CONNECT_HOST" != "https://"* ]]; then
+  export OP_CONNECT_HOST="http://"$OP_CONNECT_HOST
+fi
```
florisvdg, line 19: One of the main reasons to use the CLI in the action is to get local vs. CI parity. Adding custom environment variable logic in the action alone would be a step backwards.

Also raises the question, why `http` and not `https`? The answer is probably: we should let the user decide that and not make any assumptions here.

If we do really want to add this, we should probably do that on the CLI side and bump the CLI version in the action.

action.yml
```diff
@@ -10,7 +10,7 @@ inputs:
     default: false
   export-env:
     description: Export the secrets as environment variables
-    default: false
+    default: true
```
florisvdg, line 13: I recall that making `export-env` an opt-in (so default `false`) was a conscious decision, which I don't think we should just revert.

florisvdg (commented):
~Isn't the real solution here to make the CLI-based action `v2`? And keep the `curl`-based `v1`?~

Given that there have already been `v1.1.x` tags containing these changes, the path forward would be to create a `v1.1.x` patch to undo the breaking changes. The changes can be added again in a `v2` release.

README.md
```diff
@@ -173,8 +171,6 @@ jobs:
       - name: Load AWS credentials
         uses: 1password/load-secrets-action@v1
         with:
-          # Export loaded secrets as environment variables
-          export-env: true
```
florisvdg, line 177: Could be good to still keep these in the README for documentation purposes.

florisvdg (approved):

README.md
edif2008: Sure, will keep them in the docs then.

PR #28, opened 2023-01-29
Use temp directory for `op` (versus static `/usr/local/bin`)

### LINKED ISSUE #26: Permission denied when using self-hosted
When using the self-hosted version, specifically using kubernetes and the summerwind image (summerwind/actions-runner-dind:ubuntu-20.04) I got the following:

```
Authenticated with CONNECT 
Archive:  op.zip
error:  cannot create /usr/local/bin/op.sig
        Permission denied
error:  cannot create /usr/local/bin/op
        Permission denied
Error: The process '/usr/bin/sh' failed with exit code 50
```
Since it tries to decompress the file in that directory and that directory is owned by root, it throws an error. The pod runs with user "runner"

Any thoughts?

mamercad:
I ran into the same with with `self-hosted`; switch from `/usr/local/bin` to a temporary directory (which is created and removed instead); fixes #26.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -39,12 +39,26 @@ unset_prev_secrets() {
 
 # Install op-cli
 install_op_cli() {
+  OP_INSTALL_DIR="$(mktemp -d)"
+  if [[ ! -d "$OP_INSTALL_DIR" ]]; then
+    echo "Install dir $OP_INSTALL_DIR not found"
+    exit 1
+  fi
+  export OP_INSTALL_DIR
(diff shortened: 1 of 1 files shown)


utamas:
Hi @edif2008 (sorry, for randomly picking you from contributors). Any chance this PR will be merged any time soon (or some other solution to permission denied problem we face)?

edif2008 (approved):
Thank you so much for the contribution @mamercad. 😊 
We really appreciate you making the action a better integration that works in more use cases.

Issue #27, opened 2023-01-27
GitHub Actions `set-output` is deprecated and scheduled for removal


coreyti:
Hi,

This action currently relies upon the `echo "::set-output ..."` functionality of GitHub actions. However, that feature is [now deprecated and scheduled to be removed](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) at the end of May.

We've just started integrating this 1Password Connect workflow into our pipelines, and plan to be heavy users of the (wonderful) integration. So, hoping things don't break down the line.

Thanks!

dustin-ruetz:
Quick update on this issue: we are planning to remove `::set-output` from the codebase relatively soon, but in the meantime the [GitHub blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) has been updated stating that they are postponing its removal due to significant usage and the large number customers that would be impacted.

<img width="761" alt="image" src="https://github.com/1Password/load-secrets-action/assets/26169612/dee88a48-ded9-4cf4-94f4-8de74d7e3758">

PR #29, opened 2023-02-13
Use environment file to manage output instead of `set-output`

Manokii:
`set-output` & `save-state` are now deprecated

resolves 1Password/load-secrets-action#27


edif2008:
Thank you so much for your contribution on this. 
I really appreciate you making sure that we no longer use deprecated syntax. 😄

edif2008 (approved):

jillianwilson (approved):

PR #34, opened 2023-03-22
Tooling improvements

dustin-ruetz:
## Summary

* This PR is intended as the initial step towards migrating this GitHub Action to a fully-TypeScript codebase.
* It focuses on implementing tooling to help give us confidence in the future code changes that we plan to make.

## Features

* It switches building the code from CommonJS to ESModules.
* It updates the lint.yml config as per the most recent [ShellCheck README](https://github.com/ludeeus/action-shellcheck#ignore-paths-and-names) file.
* It updates dependencies and devDependencies to their latest versions.
  * The notable exception to this is the TypeScript package, and the reason why this one was not updated is because migrating from v4 to v5 produces the below error. Checking the https://www.typescriptlang.org/tsconfig reference shows that they haven’t added the `verbatimModuleSyntax` option/usage examples yet, presumably because TypeScript v5 was released on 2023-03-16 (i.e. less than one week ago as of this writing).
  * My recommendation is: 1) we don't implement TypeScript's `"ignoreDeprecations": "5.0"` suggestion, and 2) we upgrade to v5 after the TS docs have been updated to include the `verbatimModuleSyntax` option and usage examples.

  ```
  Option 'importsNotUsedAsValues' is deprecated and will stop functioning in TypeScript 5.5. Specify compilerOption '"ignoreDeprecations": "5.0"' to silence this error.
  Use 'verbatimModuleSyntax' instead.
  ```

* It adds ESLint for code linting and Prettier for code formatting. It also adds the @1password/front-end-style package and and uses its `eslintrc.yml` and `prettierrc.json` config files with the aim of ensuring consistent linting/formatting across all of our 1Password codebases.
  * This repo is a GitHub Action and so the "front-end-style" part of the package name is a misnomer, but the conventions that the package offers are very relevant to our JS/TS files.
* It adds Jest for unit testing our TS files.
* It adds an alphabetized list of NPM scripts, most of which are extendable (example:  `lint:fix` is built on top of `lint`.)
* It adds [Husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged) to automatically trigger pre-commit and pre-push checks.


tsconfig.json
```diff
@@ -1,13 +1,25 @@
...
+		"moduleResolution": "node",
+		"noEmit": true,
+		"noEmitOnError": true,
+		"noFallthroughCasesInSwitch": true,
+		"noImplicitReturns": true,
+		"noUncheckedIndexedAccess": true,
+		"noUnusedLocals": true,
+		"noUnusedParameters": true,
+		"outDir": "dist/",
+		"rootDir": "src/",
```
edif2008, line 21: Comparing to the previous version of this, we had the `./` prefix to suggest to look for the current directory.

I'm asking this as a sanity check, but does removing this have any side effects? Are there any chances that with this config, the project to look at the wrong place?

src/index.ts
```diff
@@ -1,21 +1,28 @@
...
+		let message = "Unknown Error";
+		if (error instanceof Error) {
+			message = error.message;
+		}
+		core.setFailed(message);
+	}
+};
 
-run();
+void run();
```
edif2008, line 28: I wonder if there's a real difference between `run()` and `void run()` in this case. 🤔

src/index.ts
```diff
@@ -1,21 +1,28 @@
...
+		process.env.INPUT_EXPORT_ENV = core.getInput("export-env");
 
-  } catch (error: any) {
-    core.setFailed(error.message);
-  }
-}
+		// Execute bash script
+		await exec.exec(`sh -c "` + parentDir + `/entrypoint.sh"`);
+	} catch (error) {
+		// https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript
```
edif2008, line 19: Maybe a silly nit, but adding some context for this link may help a bit.
The link itself may be explanatory, but since it's not from some official documentation (e.g. https://www.typescriptlang.org/docs/) I may want to know why we use this source as a reference.

package.json
```diff
@@ -1,37 +1,68 @@
...
-    "@actions/core": "^1.10.0",
-    "@actions/exec": "^1.1.1"
-  },
-  "devDependencies": {
-    "@types/node": "^18.14.2",
-    "@vercel/ncc": "^0.36.1",
-    "typescript": "^4.9.5"
-  }
+	"name": "load-secrets-action",
+	"version": "1.1.0",
```
edif2008, line 3: Since we're doing some tooling improvement here, would it be ok to include it in this PR making this version match the action's version? (currently `1.2.0`)

package.json
```diff
@@ -1,37 +1,68 @@
...
+		"1password",
+		"load secrets",
+		"connect"
+	],
+	"author": "1Password",
+	"license": "MIT",
+	"bugs": {
+		"url": "https://github.com/1Password/load-secrets-action/issues"
+	},
+	"homepage": "https://github.com/1Password/load-secrets-action#readme",
```
edif2008, line 42: I wonder if it would be better to link here to the github action documentation on developer.1password.com.
Specifically, this one: https://developer.1password.com/docs/ci-cd/github-actions

dist/package.json
edif2008, line 1: Any reason for adding this file here? 🤔

action.yml
```diff
@@ -12,5 +12,5 @@ inputs:
     description: Export the secrets as environment variables
     default: true
 runs:
-  using: 'node16'
-  main: 'dist/index.js'
+  using: "node16"
+  main: "dist/index.js"
```
edif2008, line 16: Just wondering if there's any preference/good practice here as to which one is better suited 🤔

.github/workflows/lint.yml
```diff
@@ -5,6 +5,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: ShellCheck
-        uses: ludeeus/action-shellcheck@1.1.0
+      - uses: actions/checkout@v3
+      - name: Run ShellCheck
+        uses: ludeeus/action-shellcheck@master
```
edif2008, line 10: While this matches with their usage examples, it now adds the risk of things breaking especially if the action will have a major release (e.g. from v2 to v3). 

Looking at it now, v2 was released around 2 months ago, while we used v1. This v2 doesn't bring any breaking changes fortunately, but other major release may bring some.

Another reason why `master` tag looks risky is that changes may be added that weren't tested or are unstable, leading to the small chance of breaking the action. 


Therefore, my question is: should we better use a version tag (in this case `2.0.0`) rather than the `master` tag?

edif2008 (commented):
Thank you so much for doing this. This will definitely improve the quality of the action and ease its maintenance on the long run. 💪🏻 

I've left a couple of comments that are mostly nits and clarification questions, but so far this looks solid to me.

One more question I have. Should we enable dependabot for this repo?

tsconfig.json
dustin-ruetz: Good question; in this case both with the `./` prefix and without it will work in the exact same way. I did use the `./` prefix consistently in the `package.json` when referencing folders and files, so it is probably good practice to do the same here in `tsconfig.json` as well.

action.yml
jodyheavener: I am neutral as long as it's consistent and enforced, which is what appears to be happening

config/.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm run pre-push
```
jodyheavener, line 4: Yay Husky

dist/package.json
jodyheavener: Do dist files need to be in the repo? I don't know how publishing works for Actions but is there not a build step to avoid putting these in the repo?

action.yml
dustin-ruetz: Changing the strings from single quotes to double quotes was performed automatically by the 1password/front-end-style Prettier config, so I'll mark this conversation as resolved.

src/index.ts
jodyheavener: I used void in these instances too, since we're not awaiting anything, there's nothing to `then`, and if it throws we want the error to blow this up

src/index.ts
dustin-ruetz: @edif2008 - Not a silly nit at all! For context, Kent C. Dodds in a well-known JavaScript developer and a trustworthy reference 🥇

I've added a short comment to add some context on what this code does in addition to the link to KCD's article. I also re-checked his article and noticed that I missed his edge case example where he addresses if something modifies the `Error` constructor, and so I've added that missing piece here too 👍

package.json
dustin-ruetz: I can bump this manually to `1.2.0` if it makes sense to do so 👍

Seeing as how we already have a mismatch between the `package.json` version and the released version, IMO we should consider automating this using something like [semantic-release](https://github.com/semantic-release/semantic-release), a tool designed to be used in CI/CD to automatically manage versioning.

package.json
dustin-ruetz: As per the [NPM docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#homepage), the package.json `homepage` field is intended for "the url to the project homepage", and the example they list there is the repo URL as well. We already link to the 1Password Developer Portal in our README.md file, which is the more appropriate place for this IMO.

dist/package.json
dustin-ruetz: @edif2008: This file was created as a result of `ncc build` and as per https://github.com/vercel/ncc#usage-1 - "If building an `.mjs` or `.js` module inside a `"type": "module"` [package boundary](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_json_and_file_extensions), an ES module output will be created automatically. [...] Note: If the input file is using a `.cjs` extension, then so will the corresponding output file. This is useful for packages that want to use `.js` files as modules in native Node.js using a `"type": "module"` in the package.json file."

@jodyheavener: As per https://github.com/actions/typescript-action#publish-to-a-distribution-branch - "Actions are run from GitHub repos so we will checkin the packed dist folder."

tsconfig.json
dustin-ruetz: Resolved via https://github.com/1Password/load-secrets-action/pull/34/commits/6fa1be2536550171da85d05f890fad7394182d76 commit.

src/index.ts
dustin-ruetz: Resolved via https://github.com/1Password/load-secrets-action/pull/34/commits/0eebc66811c3675d2436786f531f3f00d8969883 commit.

package.json
edif2008: Using an automation tool sounds like an awesome idea. However, I'd suggest investigating and providing such solution in a separate PR if that makes more sense. 

Taking a quick look at the [semantic-release](https://github.com/semantic-release/semantic-release) tool, it looks like there may be more to do on our side than just using it (e.g. commit format).

Therefore, here's my suggestion for this:
- Manually bump the version in this PR to no longer have the mismatch.
- Later add a tool to automatically manage versioning so that such mismatches no longer happen.

.github/workflows/lint.yml
edif2008: That sounds like a good idea, but unfortunately it will not work for this specific action.

According to [Using tags for release management for GitHub Actions](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management):
> Move the major version tag (such as v1, v2) to point to the Git ref of the current release.

Looking at the tags for that action, they don't have `v2` or `2`, they only have `2.0.0`. Therefore, `@2` will reference to nothing 😞

edif2008 (approved):

action.yml
jodyheavener, line 1: Unrelated to your change, but I wanted to call out that apparently using Boolean for input defaults doesn't match the GitHub Action YAML schema?

CleanShot 2023-03-27 at 09 10 49@2x

package-lock.json
jodyheavener, line 1: When I check this branch out and run `npm i` I get a diff in this file. Think you'll need to update this.

```diff
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
        "name": "load-secrets-action",
-       "version": "1.1.0",
+       "version": "1.2.0",
        "lockfileVersion": 2,
        "requires": true,
        "packages": {
                "": {
                        "name": "load-secrets-action",
-                       "version": "1.1.0",
+                       "version": "1.2.0",
                        "license": "MIT",
                        "dependencies": {
                                "@actions/core": "^1.10.0",
```

README.md
```diff
@@ -249,10 +247,10 @@ jobs:
...
 ### `configure` Action Inputs
 
-| Name                    | Default | Environment variable       | Description                                              |
-| ----------------------- | ------- | -------------------------- | -------------------------------------------------------- |
-| `connect-host`          |         | `OP_CONNECT_HOST`          | Your 1Password Connect instance URL                      |
-| `connect-token`         |         | `OP_CONNECT_TOKEN`         | Token to authenticate to your 1Password Connect instance |
+| Name            | Default | Environment variable | Description                                              |
+| --------------- | ------- | -------------------- | -------------------------------------------------------- |
+| `connect-host`  |         | `OP_CONNECT_HOST`    | Your 1Password Connect instance URL                      |
+| `connect-token` |         | `OP_CONNECT_TOKEN`   | Token to authenticate to your 1Password Connect instance |
```
jodyheavener, line 253: If there's no defaults in this table, maybe we don't need this column?

package.json
```diff
@@ -1,37 +1,68 @@
...
+		"format:write": "npm run format -- --write ./",
+		"lint": "eslint ./",
+		"lint:fix": "npm run lint -- --fix",
+		"pre-commit": "lint-staged --config ./config/lint-staged.config.js",
+		"pre-push": "npm run validate",
+		"prepare": "husky install ./config/.husky",
+		"test": "jest --config=./config/jest.config.js",
+		"test:clearcache": "jest --clearCache",
+		"test:coverage": "npm run test -- --coverage",
+		"test:watch": "npm run test -- --watch",
```
jodyheavener, line 23: I wish we were using Yarn, this would just be `yarn test --watch` :lolsob:

package.json
```diff
@@ -1,37 +1,68 @@
...
+	},
+	"scripts": {
+		"build": "ncc build ./src/index.ts",
+		"format": "prettier --ignore-path ./config/.prettierignore",
+		"format:check": "npm run format -- --check ./",
+		"format:write": "npm run format -- --write ./",
+		"lint": "eslint ./",
+		"lint:fix": "npm run lint -- --fix",
+		"pre-commit": "lint-staged --config ./config/lint-staged.config.js",
+		"pre-push": "npm run validate",
```
jodyheavener, line 18: ~Are these necessary? I thought you wouldn't need to specify these since we're registering Husky hooks in `.husky`? For example, [op-vscode uses a Husky pre-commit hook](https://github.com/1Password/op-vscode/blob/main/package.json#LL27C30-L27C30) and doesn't need to register it in scripts. What's the difference?~

package.json
```diff
@@ -1,37 +1,68 @@
...
+		"prepare": "husky install ./config/.husky",
+		"test": "jest --config=./config/jest.config.js",
+		"test:clearcache": "jest --clearCache",
+		"test:coverage": "npm run test -- --coverage",
+		"test:watch": "npm run test -- --watch",
+		"typecheck": "tsc",
+		"validate": "npm run format:check && npm run lint && npm run test:coverage && npm run typecheck && npm run build"
```
jodyheavener, line 25: This one's a doozy - if it's only getting called from the `pre-push` script, which only exists to get called from within the husky script, maybe it can just go in the husky script itself?

tsconfig.json
```diff
@@ -1,13 +1,25 @@
...
-    },
-    "exclude": ["node_modules"]
+	"compilerOptions": {
+		"allowJs": false,
+		"allowUnreachableCode": false,
+		"allowUnusedLabels": false,
+		"esModuleInterop": true,
+		"exactOptionalPropertyTypes": true,
+		"forceConsistentCasingInFileNames": true,
+		"importsNotUsedAsValues": "error",
```
jodyheavener, line 9: FYI

CleanShot 2023-03-27 at 09 33 12@2x

jodyheavener (commented):
This is great! I have a couple minor nits for ya.

tsconfig.json
dustin-ruetz: I noticed this error as well after attempting to upgrade to TypeScript 5; see below screenshot (from the "Features" section in the PR description) for more details.

image

action.yml
dustin-ruetz: Good callout 🤔 Is this a VS Code extension that lints YAML files? If yes, what's the name of the extension so that I can validate this on my end?

Also tagging @edif2008 - any ideas on this one?

package-lock.json
dustin-ruetz: Good catch, I neglected to run `npm install` after manually bumping the `package.json` version field to 1.2.0 in a previous commit.

Resolving this conversation based on the https://github.com/1Password/load-secrets-action/pull/34/commits/31a131125dd4ffaffc40e7644f26d9264f5baf33 commit.

tsconfig.json
jodyheavener: Oh yes, sorry, thanks for calling this out. I think we could ignore it for now.

package.json
dustin-ruetz: Good callout 👍 Resolving this conversation via the https://github.com/1Password/load-secrets-action/pull/34/commits/1eecfadbf27d622b4ecdcac5ec3b274dc9345fa0 commit.

jodyheavener (approved):
Excellent!

package.json
edif2008: We can switch to Yarn at a later stage if we feel like it is needed (preferably in a separate PR if we decide to do so).

Since this is a GitHub Action and not a regular TS project, the chances of us increasing dependencies and other elements is small, therefore I feel that keeping the number of additional tools to a minimum seems a better approach.

But I am open to suggestions about this one 😄

README.md
edif2008: They are undefined indeed.

The way the action works is that is you provide a value to e.g. `connect-host`, then the `configure` action will make it available for the following steps as an environment variable. The default is the empty string and in this case, the action will do nothing.

Therefore, we can safely remove the `Default` column since it brings no additional value.

action.yml
edif2008: I also get this warning on VS Code.

Looking at the documentation regarding [GitHub Action inputs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs), it looks like their example for boolean values they provide the value without the quotes.

However, reading the [porperty's description](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault):
> A `string` representing the default value. The default value is used when an input parameter isn't specified in a workflow file.

Therefore, I feel like adding quotes to it should make both parts happy. 
They seem to be using the `'` quotes instead of `"`, but I will leave that to you as in which is better. 🤔

dustin-ruetz: Resolving this conversation as per the https://github.com/1Password/load-secrets-action/pull/34/commits/1583046e3e60d22f35d460972e7202ff6827cd9b commit.

PR #36, opened 2023-04-10
Draft: Migrate action to Typescript

edif2008:
This PR migrates most of the functionality of the GitHub Action to Typescript.

## Key elements used in the migration
| Element | Explanation |
|:---------|:-----------|
| [`op-js`](https://github.com/1Password/op-js) | Typescript package that wraps 1Password CLI commands. It enables us to easily work with the output returned by the 1Password CLI. |
| [`core.setOutput`](https://github.com/actions/toolkit/tree/master/packages/core#inputsoutputs) | Make the secret available as a step's output (works with multiline secrets as well). |
| [`core.exportVariable`](https://github.com/actions/toolkit/tree/master/packages/core#exporting-variables) | Make the secret available as an environment variable (works with multiline secrets as well). |
| [`core.setSecret`](https://github.com/actions/toolkit/tree/master/packages/core#setting-a-secret) | Ensures the secret is masked in logs. |

## Migration process
The commits should show an overview of the steps of the migration. I will present here the details of the main migration step highlighted in [`6c9a28c`](https://github.com/1Password/load-secrets-action/commit/6c9a28c6b2e4fc30bb9c9302a7878a89bf8bbd47):
| Function from `entrypoint.sh` | Functionality in `index.ts` |
|--------|-------|
| auth validation (no dedicated function) | `validateAuth` |
| `install_cli` | `installCLI` |
| `extract_secrets`. `populating_secret` | `extractSecrets` |

Note: `installCLI` in `index.tx` *only* executes the shell script that will perform the CLI installation. No real migration has been performed for this part of the action.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,33 +1,5 @@
 #!/bin/bash
-# shellcheck disable=SC2046,SC2001,SC2086
 set -e
-
-# Pass User-Agent Inforomation to the 1Password CLI
-export OP_INTEGRATION_NAME="1Password GitHub Action"
-export OP_INTEGRATION_ID="GHA"
-export OP_INTEGRATION_BUILDNUMBER="1010001"
-
-managed_variables_var="OP_MANAGED_VARIABLES"
-IFS=','
-
-# Unset all secrets managed by 1Password if `unset-previous` is set.
-unset_prev_secrets() {
-  if [ "$INPUT_UNSET_PREVIOUS" == "true" ]; then
-    echo "Unsetting previous values..."
-
-    # Find environment variables that are managed by 1Password.
-    for env_var in "${managed_variables[@]}"; do
-      echo "Unsetting $env_var"
-      unset $env_var
-
-      echo "$env_var=" >> $GITHUB_ENV
-
-      # Keep the masks, just in case.
-    done
-
-    managed_variables=()
-  fi
-}
 
 # Install op-cli
 install_op_cli() {
@@ -47,6 +19,7 @@ install_op_cli() {
     tar -xvf temp-pkg/op.pkg/Payload -C "$OP_INSTALL_DIR"
     rm -rf temp-pkg && rm op.pkg
   fi
+  echo "$OP_INSTALL_DIR" >> "$GITHUB_PATH"
 }
 
 # Uninstall op-cli
@@ -56,72 +29,4 @@ uninstall_op_cli() {
   fi
 }
 
-populating_secret() {
-  ref=$(printenv $1)
-
-  echo "Populating variable: $1"
-  secret_value=$("${OP_INSTALL_DIR}/op" read "$ref")
-
-  if [ -z "$secret_value" ]; then
-    echo "Could not find or access secret $ref"
-    exit 1
-  fi
-
-  # Register a mask for the secret to prevent accidental log exposure.
-  # To support multiline secrets, escape percent signs and add a mask per line.
-  escaped_mask_value=$(echo "$secret_value" | sed -e 's/%/%25/g')
-  IFS=$'\n'
-  for line in $escaped_mask_value; do
-    if [ "${#line}" -lt 3 ]; then
-      # To avoid false positives and unreadable logs, omit mask for lines that are too short.
-      continue
-    fi
-    echo "::add-mask::$line"
-  done
-  unset IFS
-
-  if [ "$INPUT_EXPORT_ENV" == "true" ]; then
-    # To support multiline secrets, we'll use the heredoc syntax to populate the environment variables.
-    # As the heredoc identifier, we'll use a randomly generated 64-character string,
-    # so that collisions are practically impossible.
-    random_heredoc_identifier=$(openssl rand -hex 32)
-
-    {
-      # Populate env var, using heredoc syntax with generated identifier
-      echo "$env_var<<${random_heredoc_identifier}"
-      echo "$secret_value"
-      echo "${random_heredoc_identifier}"
-    } >> $GITHUB_ENV
-    echo "GITHUB_ENV: $(cat $GITHUB_ENV)"
-
-  else
-    # Prepare the secret_value to be outputed properly (especially multiline secrets)
-    secret_value=$(echo "$secret_value" | awk -v ORS='%0A' '1')
-
-    echo "::set-output name=$env_var::$secret_value"
-  fi
-
-  managed_variables+=("$env_var")
-}
-
-# Load environment variables using op cli. Iterate over them to find 1Password references, load the secret values,
-# and make them available as environment variables in the next steps.
-extract_secrets() {
-  IFS=$'\n'
-  for env_var in $("${OP_INSTALL_DIR}/op" env ls); do
-    populating_secret $env_var
-  done
-}
-
-read -r -a managed_variables <<< "$(printenv $managed_variables_var)"
-
-unset_prev_secrets
 install_op_cli
-extract_secrets
-uninstall_op_cli
-
-unset IFS
-# Add extra env var that lists which secrets are managed by 1Password so that in a later step
-# these can be unset again.
-managed_variables_str=$(IFS=','; echo "${managed_variables[*]}")
-echo "$managed_variables_var=$managed_variables_str" >> $GITHUB_ENV
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -7,16 +7,8 @@ export OP_INTEGRATION_NAME="1Password GitHub Action"
 export OP_INTEGRATION_ID="GHA"
 export OP_INTEGRATION_BUILDNUMBER="1010001"
 
-readonly CONNECT="CONNECT"
-readonly SERVICE_ACCOUNT="SERVICE_ACCOUNT"
-
-auth_type=$CONNECT
 managed_variables_var="OP_MANAGED_VARIABLES"
 IFS=','
-
-if [[ "$OP_CONNECT_HOST" != "http://"* ]] && [[ "$OP_CONNECT_HOST" != "https://"* ]]; then
-  export OP_CONNECT_HOST="http://"$OP_CONNECT_HOST
-fi
 
 # Unset all secrets managed by 1Password if `unset-previous` is set.
 unset_prev_secrets() {
@@ -122,17 +114,6 @@ extract_secrets() {
 }
 
 read -r -a managed_variables <<< "$(printenv $managed_variables_var)"
-
-if [ -z "$OP_CONNECT_TOKEN" ] || [ -z "$OP_CONNECT_HOST" ]; then
-  if [ -z "$OP_SERVICE_ACCOUNT_TOKEN" ]; then
-    echo "(\$OP_CONNECT_TOKEN and \$OP_CONNECT_HOST) or \$OP_SERVICE_ACCOUNT_TOKEN must be set"
-    exit 1
-  fi
-
-  auth_type=$SERVICE_ACCOUNT
-fi
-
-printf "Authenticated with %s \n" $auth_type
 
 unset_prev_secrets
 install_op_cli
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,16 +5,13 @@ import * as exec from "@actions/exec";
 
 const run = async () => {
 	try {
-		const currentFile = url.fileURLToPath(import.meta.url);
-		const currentDir = path.dirname(currentFile);
-		const parentDir = path.resolve(currentDir, "..");
 
 		// Get action inputs
 		process.env.INPUT_UNSET_PREVIOUS = core.getInput("unset-previous");
 		process.env.INPUT_EXPORT_ENV = core.getInput("export-env");
 
 		// Execute bash script
-		await exec.exec(`sh -c "` + parentDir + `/entrypoint.sh"`);
+		await executeScript();
 	} catch (error) {
 		// It's possible for the Error constructor to be modified to be anything
 		// in JavaScript, so the following code accounts for this possibility.
@@ -28,5 +25,14 @@ const run = async () => {
 		core.setFailed(message);
 	}
 };
+
+const executeScript = async (): Promise<void> => {
+	const currentFile = url.fileURLToPath(import.meta.url);
+	const currentDir = path.dirname(currentFile);
+	const parentDir = path.resolve(currentDir, "..");
+
+	// Execute bash script
+	await exec.exec(`sh -c "` + parentDir + `/entrypoint.sh"`);
+}
 
 void run();
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import path from "path";
 import url from "url";
 import * as core from "@actions/core";
 import * as exec from "@actions/exec";
+import { read, setClientInfo } from "@1password/op-js";
 
 const envConnectHost = "OP_CONNECT_HOST";
 const envConnectToken = "OP_CONNECT_TOKEN";
@@ -12,12 +13,24 @@ const run = async () => {
 		// Validate that a proper authentication configuration is set for the CLI
 		validateAuth();
 
-		// Get action inputs
-		process.env.INPUT_UNSET_PREVIOUS = core.getInput("unset-previous");
-		process.env.INPUT_EXPORT_ENV = core.getInput("export-env");
+		// Download and install the CLI
+		await installCLI();
 
-		// Execute bash script
-		await executeScript();
+		// Get action inputs
+		const unsetPrevious = core.getBooleanInput("unset-previous");
+		const exportEnv = core.getBooleanInput("export-env");
+
+		// Unset all secrets managed by 1Password if `unset-previous` is set.
+		if (unsetPrevious && process.env.OP_MANAGED_VARIABLES) {
+			core.debug(`Unsetting previous values ...`);
+			const managedEnvs = process.env.OP_MANAGED_VARIABLES.split(",");
+			for (const envName of managedEnvs) {
+				core.debug(`Unsetting ${envName}`);
+				core.exportVariable(envName, "");
+			}
+		}
+
+		await extractSecrets(exportEnv);
 	} catch (error) {
 		// It's possible for the Error constructor to be modified to be anything
 		// in JavaScript, so the following code accounts for this possibility.
@@ -56,13 +69,47 @@ const validateAuth = () => {
 	core.debug(`Authenticated with ${authType}.`);
 };
 
-const executeScript = async (): Promise<void> => {
+/* eslint-disable @typescript-eslint/naming-convention */
+const installCLI = async (): Promise<void> => {
 	const currentFile = url.fileURLToPath(import.meta.url);
 	const currentDir = path.dirname(currentFile);
 	const parentDir = path.resolve(currentDir, "..");
 
 	// Execute bash script
 	await exec.exec(`sh -c "` + parentDir + `/entrypoint.sh"`);
+};
+
+const extractSecrets = async (exportEnv: boolean) => {
+	// Pass User-Agent Inforomation to the 1Password CLI
+	setClientInfo({
+		name: "1Password GitHub Action",
+		id: "GHA",
+		build: "1020000",
+	});
+
+	// Load environment variables using 1Password CLI. Iterate over them to find 1Password references,
+	// load the secret values, and make them available either as step output or as environment variables
+	// in the next steps.
+	const res = await exec.getExecOutput(`sh -c "op env ls"`);
+	const envs = res.stdout.replace(/\n+$/g, "").split(/\r?\n/);
+	for (const envName of envs) {
+		core.debug(`Populating variable: ${envName}`);
+		const ref = process.env[envName];
+		if (ref) {
+			const secretValue = read.parse(ref);
+			if (secretValue) {
+				if (exportEnv) {
+					core.exportVariable(envName, secretValue);
+				} else {
+					core.setOutput(envName, secretValue);
+				}
+				core.setSecret(secretValue);
+			}
+		}
+	}
+	if (exportEnv) {
+		core.exportVariable("OP_MANAGED_VARIABLES", envs.join());
+	}
 };
 
 void run();
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,8 +3,14 @@ import url from "url";
 import * as core from "@actions/core";
 import * as exec from "@actions/exec";
 
+const envConnectHost = "OP_CONNECT_HOST";
+const envConnectToken = "OP_CONNECT_TOKEN";
+const envServiceAccountToken = "OP_SERVICE_ACCOUNT_TOKEN";
+
 const run = async () => {
 	try {
+		// Validate that a proper authentication configuration is set for the CLI
+		validateAuth();
 
 		// Get action inputs
 		process.env.INPUT_UNSET_PREVIOUS = core.getInput("unset-previous");
@@ -26,6 +32,30 @@ const run = async () => {
 	}
 };
 
+const validateAuth = () => {
+	let authType = "Connect";
+	if (!process.env[envConnectHost] || !process.env[envConnectToken]) {
+		if (!process.env[envServiceAccountToken]) {
+			throw new Error(
+				`(${envConnectHost} and ${envConnectToken}) or ${envServiceAccountToken} must be set`,
+			);
+		}
+		authType = "Service account";
+	}
+
+	// Adjust Connect host to have a protocol
+	if (
+		process.env[envConnectHost] &&
+		/* eslint-disable no-restricted-syntax */
+		(!process.env[envConnectHost]?.startsWith("http://") ||
+			!process.env[envConnectHost].startsWith("https://"))
+	) {
+		process.env[envConnectHost] = `http://${process.env[envConnectHost]}`;
+	}
+
+	core.debug(`Authenticated with ${authType}.`);
+};
+
 const executeScript = async (): Promise<void> => {
 	const currentFile = url.fileURLToPath(import.meta.url);
 	const currentDir = path.dirname(currentFile);
@@ -33,6 +63,6 @@ const executeScript = async (): Promise<void> => {
 
 	// Execute bash script
 	await exec.exec(`sh -c "` + parentDir + `/entrypoint.sh"`);
-}
+};
 
 void run();
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
 			"version": "1.2.0",
 			"license": "MIT",
 			"dependencies": {
+				"@1password/op-js": "^0.1.8",
 				"@actions/core": "^1.10.0",
 				"@actions/exec": "^1.1.1"
 			},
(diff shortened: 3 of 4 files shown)


src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const envConnectHost = "OP_CONNECT_HOST";
+export const envConnectToken = "OP_CONNECT_TOKEN";
+export const envServiceAccountToken = "OP_SERVICE_ACCOUNT_TOKEN";
+export const envManagedVariables = "OP_MANAGED_VARIABLES";
+
+export const authErr = `(${envConnectHost} and ${envConnectToken}) or ${envServiceAccountToken} must be set`;
```
dustin-ruetz, line 6: This may be a nitpick, but can we make this error clearer by making it a bit more specific? For example:

`Authorization error with environment variables: you must set either 1) ${envServiceAccountToken}, or 2) both ${envConnectHost} and ${envConnectToken}.`

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	// Adjust Connect host to have a protocol
+	if (
+		process.env[envConnectHost] &&
+		/* eslint-disable no-restricted-syntax */
+		!process.env[envConnectHost].startsWith("http://") &&
+		!process.env[envConnectHost].startsWith("https://")
+	) {
+		process.env[envConnectHost] = `http://${process.env[envConnectHost]}`;
```
dustin-ruetz, line 33: Pretty sure I already know the answer to this one, but just want to vocalize it here as a gut check: Connect host URLs are always `http` and not `https` because the Connect server is meant to be hosted inside of your own trusted network infrastructure - is that correct?

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+		.join("");
+
+export const validateAuth = (): void => {
+	let authType = "Connect";
+	if (!process.env[envConnectHost] || !process.env[envConnectToken]) {
+		if (!process.env[envServiceAccountToken]) {
+			throw new Error(authErr);
+		}
+		authType = "Service account";
+	}
```
dustin-ruetz, line 24: Was just testing out some local changes to this section of the code; we can simplify it quite a bit with the following changes:

1. Don't declare `authType`/default it to "Connect".
2. Create some boolean checks:
  ```ts
  const isConnect = process.env[envConnectHost] && process.env[envConnectToken];
  const isServiceAccount = process.env[envServiceAccountToken];
  ```
3. Use these booleans to check for issues with the environment variables:
  ```ts
  if (isConnect && isServiceAccount) {
    throw new Error("Environment variables were passed for both Connect and Service Accounts; you must choose either one or the other, but not both.");
  }
  if (!isConnect && !isServiceAccount) {
    throw new Error(authErr);
  }
  ```
4. If there are no issues with the env vars, then we can set the authType.
  ```ts
  const authType = isConnect ? "Connect" : "Service account";
  ```

Curious to hear if these changes make sense to you, especially the additional error handling that I added if someone provides all three env vars (i.e. two for Connect, one for Service Accounts).

dustin-ruetz (changes requested):

.github/workflows/test.yml
```diff
@@ -2,11 +2,21 @@ on: push
 name: Run acceptance tests
 
 jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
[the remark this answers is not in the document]
jodyheavener: ++ to Node 18, maybe we can add an nvmrc as well?

entrypoint.sh
```diff
@@ -1,56 +1,20 @@
...
-      # Keep the masks, just in case.
-    done
-
-    managed_variables=()
-  fi
-}
-
 # Install op-cli
 install_op_cli() {
+  CLI_VERSION="v2.16.0-beta.01"
```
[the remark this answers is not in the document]
jodyheavener: You could check out what powers `op update` to see how it checks for the latest CLI version. That said I don't mind hard-coding the version. Despite our best efforts we could introduce regressions or even intentionally have a breaking change in future CLI versions and break folks workflows.

entrypoint.sh
```diff
@@ -64,83 +28,4 @@ uninstall_op_cli() {
   fi
 }
```
jodyheavener, line 29: Out of scope but I'd love if all this was a separate published Action. There are lots of reasons to need to include the CLI in a Workflow (op-js being one of them) :)

entrypoint.sh
```diff
@@ -64,83 +28,4 @@ uninstall_op_cli() {
...
-    exit 1
-  fi
-
-  auth_type=$SERVICE_ACCOUNT
-fi
-
-printf "Authenticated with %s \n" $auth_type
-
-unset_prev_secrets
 install_op_cli
```
jodyheavener, line 31: If this is the only thing this shell file does it might be nice to rename it to something like `install_cli.sh`

src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const envConnectHost = "OP_CONNECT_HOST";
+export const envConnectToken = "OP_CONNECT_TOKEN";
+export const envServiceAccountToken = "OP_SERVICE_ACCOUNT_TOKEN";
+export const envManagedVariables = "OP_MANAGED_VARIABLES";
```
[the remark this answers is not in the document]
jodyheavener: I'm sort of on the fence about this convention because I see it as a relic of a time when we _needed_ to use upper-snake-case to denote constant/immutable values, but we've got `const` now (although you can still an const object's properties but that's a separate gripe) and you'll get a runtime error if you reassign it.

That said, if we do uppercase I don't see your suggestion as being terribly redundant - it still serves its purpose correctly (it's not our fault the value is also uppercase haha).

entrypoint.sh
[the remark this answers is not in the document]
jodyheavener: Question: is there a use case setting the version is an input to let the user control it?

src/utils.ts
jodyheavener: +1

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+import { read } from "@1password/op-js";
+import {
+	authErr,
+	envConnectHost,
+	envConnectToken,
+	envServiceAccountToken,
+	envManagedVariables,
+} from "./constants";
+
+export const semverToInt = (input: string): string =>
```
jodyheavener, line 11: Looks familiar 😛

src/index.ts
```diff
@@ -29,4 +43,44 @@ const run = async () => {
...
+	setClientInfo({
+		name: "1Password GitHub Action",
+		id: "GHA",
+		build: semverToInt(version),
+	});
+
+	// Load secrets from environment variables using 1Password CLI.
+	// Iterate over them to find 1Password references, extract the secret values,
+	// and make them available in the next steps either as step outputs or as environment variables.
+	const res = await exec.getExecOutput(`sh -c "op env ls"`);
```
jodyheavener, line 76: Do we need a try catch for this, or any sort of error checking?

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export const extractSecret = (
+	envName: string,
+	shouldExportEnv: boolean,
+): void => {
+	core.debug(`Populating variable: ${envName}`);
+	const ref = process.env[envName];
+	if (ref) {
+		const secretValue = read.parse(ref);
```
jodyheavener, line 46: Do we need to handle errors if the secret reference is invalid, references a non-existent field/item/vault?

jodyheavener (commented):
Took a quick pass at this, I love what you've done here so great job! I'm in agreement with Dustin's comments as well.

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+	core.debug(`Authenticated with ${authType}.`);
+};
+
+export const extractSecret = (
+	envName: string,
+	shouldExportEnv: boolean,
+): void => {
+	core.debug(`Populating variable: ${envName}`);
+	const ref = process.env[envName];
```
dustin-ruetz, line 44: (This comment is similar to @jodyheavener's below.) Is it possible that `ref` could be `undefined` here? If yes, what should be do to handle that scenario?

src/index.ts
```diff
@@ -29,4 +43,44 @@ const run = async () => {
...
+	// Add path to 1Password CLI to $PATH
+	const outArr = cmdOut.stdout.split("\n");
+	if (outArr[0] && process.env.PATH) {
+		const cliPath = outArr[0]?.replace(/^(::debug::OP_INSTALL_DIR: )/, "");
+		core.addPath(cliPath);
+	}
+};
+
+const loadSecrets = async (shouldExportEnv: boolean) => {
+	// Pass User-Agent Inforomation to the 1Password CLI
```
dustin-ruetz, line 66: Typo: change from `Inforomation` to `Information`.

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+				core.setOutput(envName, secretValue);
+			}
+			core.setSecret(secretValue);
+		}
+	}
+};
+
+export const unsetPrevious = (shouldUnsetPrevious: boolean): void => {
+	if (shouldUnsetPrevious && process.env[envManagedVariables]) {
+		core.debug(`Unsetting previous values ...`);
```
dustin-ruetz, line 60: This string doesn't reference any variables, so let's change it to use our oldest and most trustworthy friend: double quotes. (I'm actually surprised that Prettier didn't auto-fix this/that ESLint didn't yell at us about it.)

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+				core.exportVariable(envName, secretValue);
+			} else {
+				core.setOutput(envName, secretValue);
+			}
+			core.setSecret(secretValue);
+		}
+	}
+};
+
+export const unsetPrevious = (shouldUnsetPrevious: boolean): void => {
```
dustin-ruetz, line 58: IMO the `shouldUnsetPrevious: boolean` parameter is redundant here. We can simplify this function by refactoring it not to accept any arguments, and then in `src/index.ts` we would just call `unsetPrevious()` conditionally if `shouldUnsetPrevious` is `true`.

entrypoint.sh
dustin-ruetz: +1 on this suggestion 👍

entrypoint.sh
dustin-ruetz: As @edif2008 mentioned in the demo earlier this week, a separate GitHub Action to install the CLI is in the works 🚀

entrypoint.sh
dustin-ruetz, line 1: I just noticed that the `uninstall_op_cli()` function is defined at the end of this file, but it doesn't look like we call it anywhere. Is this an old relic from the previous version of the code that can now be removed?

src/utils.test.ts
dustin-ruetz, line 1: Let me begin by saying: excellent work on these unit tests! They are very easy to understand, and `npm run test:coverage` indicates that this file has 100% coverage 💯

I have a few suggestions to help simplify things a bit more though 😄

1. Instead of cluttering up the code with lots of calls to `jest.spyOn`, we can replace these with the following block of code right after the `import` statements to mock out all of the methods on the imported `core` object:

```ts
jest.mock("@actions/core", () => ({
	debug: jest.fn(),
	exportVariable: jest.fn(),
	setOutput: jest.fn(),
	setSecret: jest.fn(),
}));
```

2. This also results in changing the test code to more closely resemble how things are written in `src/utils.ts`:

```diff
- expect(mockCoreDebug).toBeCalledWith("Authenticated with Connect.");
+ expect(core.debug).toBeCalledWith("Authenticated with Connect."); 
```

3. This also allows us to remove most of the calls to `jest.clearAllMocks()` (with the exception of the one in the `extractSecret` describe block).

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+export const extractSecret = (
+	envName: string,
+	shouldExportEnv: boolean,
+): void => {
+	core.debug(`Populating variable: ${envName}`);
+	const ref = process.env[envName];
+	if (ref) {
+		const secretValue = read.parse(ref);
+		if (secretValue) {
+			if (shouldExportEnv) {
```
dustin-ruetz, line 48: I understand the chain of logic here, but nested `if` statements make my brain hurt :/ We have three in a row here: `if (ref)`, `if (secretValue)`, and `if (shouldExportEnv)`.

Is there a way that we can simplify this code by avoiding all of these nested `if`s?

dustin-ruetz:
Overall comment: awesome work on this TypeScript migration, @edif2008! 👏

Everything looks very solid and works well from my testing. I left a bunch of file-specific feedback in my review yesterday and some additional comments today.

Just a few remaining general thoughts:

1. `src/utils.ts` currently has four functions in it, so I'd be in favour of splitting these out into their own files in a `src/utils/` directory. They could each then get their own test suites too under `src/utils/__tests__/*.test.ts`. IMO this change would make future maintenance/refactoring easier.
2. Can we split out `installCLI` and `loadSecrets` into their own files in `src/utils/` as well?
3. This would leave `src/index.ts` with just the `run` function in it. Is there a clearer name that we can use for this function? Perhaps even just something like `loadSecretsAction`, since it's representative of the main functionality for this package?
4. It'd be great if we could add unit tests for `installCLI`, `loadSecrets` and `run` as well.

Happy to help wherever I can with the above list, especially the unit tests.

To conclude this review - once again, great work on this PR!

.github/workflows/test.yml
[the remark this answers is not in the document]
dustin-ruetz: As an FYI, [Node.js v20](https://nodejs.org/en/blog/announcements/v20-release-announce) was just released today 🚀 The "Call to action!" section at the end of that page recommends that we use either v18 or v20:

> Also of note is that Node.js 14 will go End-of-Life in April 2023, so we advise you to start planning to upgrade to Node.js 18 (LTS) or Node.js 20 (soon to be LTS).

> Please, consider that Node.js 16 (LTS) will go End-of-Life in September 2023, which was brought forward from April 2024 to coincide with the end of support of OpenSSL 1.1.1. You can read more details about that decision at https://nodejs.org/en/blog/announcements/nodejs16-eol/.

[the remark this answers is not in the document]
edif2008: Is anything stopping us from using v20 right ahead?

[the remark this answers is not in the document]
edif2008: The reason I used 16 is because the action itself uses v16. In the `action.yml` file we say the following:
```yml
runs:
  using: 'node16'
```

I'm up for updating this as well, but I couldn't find anywhere the available strings that mention the supported Node versions when creating a Node action.

I suppose changing it to `node18` or `node20` should do the trick, but I'm not sure. 🤔 

Regardless, what I want to say is that if we decide to switch it here, we should switch it everywhere.

entrypoint.sh
edif2008: I think it can now be removed, especially since I will improve the flow of installing the CLI in the following way:
- Check if the pipeline has the CLI installed
- If not, run the installation script

The nice thing is that `op-js` comes with that check, so there's very little work on our side to have this enhancement.

src/utils.ts
edif2008: Indeed, double quotes should do the trick. I wonder if there's a ESLint rule that would complain about it so that we can avoid this happening in the future. 🤔

src/constants.ts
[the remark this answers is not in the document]
edif2008: I'd prefer to keep these constants camelCase. For some reason, I feel like it looks better this way (may be because of Go). Also, Writing constant with capitals may lead to more errors? Not sure about that one.

If you feel better about having these constants use the `UPPERCASE_VARIABLES` format, then I'm up for it.

src/utils.ts
edif2008: This definitely looks way better and cleaner.

The only part of the code that I'm slightly hesitant to include is the following:
```ts
if (isConnect && isServiceAccount) {
  throw new Error("Environment variables were passed for both Connect and Service Accounts; you must choose either one or the other, but not both.");
}
```

The reason is that the CLI can handle this edge case by having Connect take priority. We count on this behavior in multiple places and we also [let the users know](https://developer.1password.com/docs/ci-cd/github-actions) that if this happens, then Connect takes priority.

It definitely feels better to stop the user from continuing when having two configurations since it can get confusing, however it may lead to a slight anti-pattern to what we do in other integrations. 🤔 
This can also be looked the other way around as in we should update all integrations to just stop if both are configured. 😄

entrypoint.sh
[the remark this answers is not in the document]
edif2008: > Is there any way we could get the most recent 1Password CLI version number programmatically?
Yes, there is, and I am switching to that one in this PR since it's one line of code 😄 

> You could check out what powers `op update` to see how it checks for the latest CLI version.
This was a great suggestion which lead to the elegant solution for this.

> That said I don't mind hard-coding the version. Despite our best efforts we could introduce regressions or even intentionally have a breaking change in future CLI versions and break folks workflows.
I'm planning on working on a separate GitHub Action that will install the CLI. In that one, getting the latest version will be the default. Users can specify which version they want to use if they want to manually update the CLI in their pipelines.

> Is there a use case setting the version as an input to let the user control it?
Yes, there is. The simplest example is ensuring that no surprises happen in the pipeline by accidentally getting the latest version of the CLI that turns out to break a pipeline. Another use case is to make sure that the CLI behaves as expected. We may improve things in future releases that may change some behavior (for the better) and some folks may want to double check that it doesn't affect their pipelines before updaing the CLI.

In the install CLi GitHub Action that I'm working on, folks will be able to specify the version they want to use. This comes in handy especially if they want to access beta features (which is another use case).

src/utils.ts
edif2008: What I did here as a safe approach is to mirror the behavior in the shell script in Typescript. Now that we're confident it was properly migrated, we can simplify it. Also, I've realized that it became like this due to the linters complaining about `undefined` and such. 

Addressed in [`7c957c5`](https://github.com/1Password/load-secrets-action/pull/36/commits/7c957c570233d0664f9b7e566485c77163b11108).

entrypoint.sh
edif2008: Addressed in [`17a2a72`](https://github.com/1Password/load-secrets-action/pull/36/commits/17a2a72e87dddbb35f43825ef9f56a930693f654).

entrypoint.sh
edif2008: Addressed in [`17a2a72`](https://github.com/1Password/load-secrets-action/pull/36/commits/17a2a72e87dddbb35f43825ef9f56a930693f654).

entrypoint.sh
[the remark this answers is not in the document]
edif2008: And the change that will programatically get the latest version of the CLI is found in commit [`17a2a72`](https://github.com/1Password/load-secrets-action/pull/36/commits/17a2a72e87dddbb35f43825ef9f56a930693f654).

src/index.ts
```diff
@@ -29,4 +43,44 @@ const run = async () => {
 	}
 };
 
+/* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
edif2008: Addressed in [`d16183d`](https://github.com/1Password/load-secrets-action/pull/36/commits/d16183d989ad9f763a5e6a67735d0347979298b5).

src/utils.test.ts
```diff
@@ -0,0 +1,134 @@
...
+	it("should throw an error when partial Connect config is provided", () => {
+		process.env[envConnectHost] = testConnectHost;
+		expect(validateAuth).toThrowError(authErr);
+	});
+
+	it("should append protocol if Connect host doesn't have it", () => {
+		process.env[envConnectHost] = "localhost:8080";
+		process.env[envConnectToken] = "token";
+		expect(validateAuth).not.toThrowError(authErr);
+		/* eslint-disable no-restricted-syntax */
```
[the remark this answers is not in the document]
edif2008: Addressed in [`d16183d`](https://github.com/1Password/load-secrets-action/pull/36/commits/d16183d989ad9f763a5e6a67735d0347979298b5).

src/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+		if (!process.env[envServiceAccountToken]) {
+			throw new Error(authErr);
+		}
+		authType = "Service account";
+	}
+
+	// Adjust Connect host to have a protocol
+	if (
+		process.env[envConnectHost] &&
+		/* eslint-disable no-restricted-syntax */
```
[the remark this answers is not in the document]
edif2008: Thank you for sharing these facts. TIL.

They will come in handy in my other TS related work, so much appreciated. 😄 

Addressed in [`d16183d`](https://github.com/1Password/load-secrets-action/pull/36/commits/d16183d989ad9f763a5e6a67735d0347979298b5).

src/utils.ts
edif2008: Definitely sounds like a cleaner approach.

Addressed in [`6ecd5ce`](https://github.com/1Password/load-secrets-action/pull/36/commits/6ecd5ceaf74eb507deb9fb2d36c18b03422431bc).

src/utils.ts
edif2008: In the meantime, I've addressed it in [`6ecd5ce`](https://github.com/1Password/load-secrets-action/pull/36/commits/6ecd5ceaf74eb507deb9fb2d36c18b03422431bc).

src/index.ts
edif2008: Addressed in [`6ecd5ce`](https://github.com/1Password/load-secrets-action/pull/36/commits/6ecd5ceaf74eb507deb9fb2d36c18b03422431bc).

config/jest.config.js
```diff
@@ -11,7 +11,13 @@ const jestConfig = {
 	testEnvironment: "node",
 	testRegex: "(/__tests__/.*|(\\.|/)test)\\.ts",
 	transform: {
-		".ts": ["ts-jest"],
+		".ts": [
+			"ts-jest",
+			{
```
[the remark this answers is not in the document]
edif2008: Totally agree with this. Probably forgot to add it after I tested it to see if it works.

Addressed in [`6ecd5ce`](https://github.com/1Password/load-secrets-action/pull/36/commits/6ecd5ceaf74eb507deb9fb2d36c18b03422431bc).

src/utils.test.ts
edif2008: This sounds like a very good idea.
I've tried this for a bit and it seems that it's not as straightforward as I'd like.

Simplest example is that if I mock `exportVariable` to an empty function, then the tests for `unsetPrevious` fill fail 🤔 
Maybe that has to do with the fact that it's recommended to not use the actual functionality of the mocked function?

Anyways, This may also no longer be the case if we split the tests into separate files, as suggested in [this comment](https://github.com/1Password/load-secrets-action/pull/36#issuecomment-1509093018). We should discuss this to make sure we're on the same page and we're going in a direction we're both happy with. 😄

src/index.ts
edif2008: It is very unlikely for this command to fail. All it does is find the environment variables that have a secret reference as a value. If there are none found, that's also fine. 😄

src/utils.ts
edif2008: I believe that the errors shown by the CLI are good enough for this case. 
Do you think we should do additional work here on top of what `op-js` is doing?

src/constants.ts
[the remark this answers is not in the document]
dustin-ruetz: I'm OK to leave them as-is in camcelCase - as Jody mentioned before the `const` is enough indication that these are immutable constant values. (Also, we have the `constants.ts` filename as an extra reminder 😄)

src/utils.ts
dustin-ruetz: Great context, thanks for expanding on the reasoning for this! I'm almost always in favour of encouraging people to be explicit with their expectations; asking users to specify either `http` or `https` as their protocol seems perfectly reasonable.

src/utils.ts
edif2008: Opened this PR to make it available directly from `op-js` (and thus no longer duplicating the code here):
 [Export semverToInt](https://github.com/1Password/op-js/pull/126).

src/utils.test.ts
edif2008: After some help and guidance from you, we've managed to simplify the code related to mocking.
This is highlighted in [`1dd370e`](https://github.com/1Password/load-secrets-action/pull/36/commits/1dd370e8b6636fb39071c08b924d06dc02a3de25).

src/utils.ts
jodyheavener: 🙌

jodyheavener: Your PR has been merged and a new version of op-js has been [published](https://www.npmjs.com/package/@1password/op-js). Update to 0.1.9 to use the new semverToInt utility :)

jodyheavener (approved):
Awesome, Eddy! This looks so good.

.github/workflows/test.yml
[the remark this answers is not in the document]
dustin-ruetz: Agreed, IMO we can close this thread now. As noted [here](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-using-nodejs-v16) (and elsewhere as we searched for examples), `node16` is currently the recommended version for actions.

dustin-ruetz (approved):

kylekurz:
Any chance this can get merged and an update to this action released? We'd like to incorporate it into some of our pipelines, but would like to have it pull in a modern build instead of the 2.10 that is in the current "live" action.

src/utils.ts
dustin-ruetz: @edif2008 Unresolving this thread because it we still need to:

1. Upgrade to the latest version of `op-js`.
2. Remove the `semverToInt` function + unit tests and replace it with the imported one from `op-js`.

src/index.ts
```diff
@@ -29,4 +38,27 @@ const run = async () => {
 	}
 };
 
-void run();
+// This function's name is an exception from the naming convention
+// since we refer to the 1Password CLI here.
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const installCLI = async (): Promise<void> => {
+	await validateCli().catch(async () => {
```
dustin-ruetz, line 45: @edif2008 @jodyheavener Taking another look at this, I'm a bit confused by the expected behaviour here. We await the promise result of `validateCli` (imported from `op-js`), but we don't handle the result in a `.then()`; instead we `.catch()` the thrown error, and then we proceed to install the CLI?

Is this expected? Maybe I am misunderstanding how this is intended to work, but after a few times of re-reading this I'm unclear as to 1) why we expect `validateCli` to throw an error, 2) why we then do a bunch of stuff inside the error's `catch` block, and 3) why we don't do anything else post-`catch`.

edif2008: We use `validateCLI` to see if there's an installed 1Password CLI in the runner. If there is, we do nothing (i.e. there's no `then()` needed). If there is no CLI installed (which will throw an error), then we run the script to install it (which is done in the `catch()` part).

I will add an additional comment on this line to explain the flow for better clarity.

src/utils.ts
edif2008: After some more discussions, we've decided that we will not add this additional enhancement on top of the CLI. Instead, we will print a warning to the user in the pipeline that both are configured and that Connect will take priority.
```ts
if (isConnect && isServiceAccount) {
	core.warning(
		"WARNING: Both service account and Connect credentials are provided. Connect credentials will take priority.",
	);
}
```

src/utils.ts
edif2008: Solved with [57ddba0](https://github.com/1Password/load-secrets-action/pull/36/commits/57ddba0a994c9e199fa23491b9239fe3005e695a) and [35650f8](https://github.com/1Password/load-secrets-action/pull/36/commits/35650f8988cd23393a4e01663ccc6cd677368432).

src/index.ts
edif2008: Added with [2c980ca](https://github.com/1Password/load-secrets-action/pull/36/commits/2c980caffa1fb9062e8ea51dc82f36f0e0150393).

src/utils.ts
edif2008: Adjusted with [27cd56a](https://github.com/1Password/load-secrets-action/pull/36/commits/27cd56abb03b44c2c565eedae6c42a8446c6cde5).

src/index.ts
dustin-ruetz: I see, makes a lot more sense to me now. Thanks for the explanation + comment!

PR #40, opened 2023-04-27
Add item create command workaround

edif2008:
This content will easily let the users know the alternate way in which they have to create items in GitHub pipelines.


florisvdg:
I'd opt for not (ab)using the README for documentation which we already have the docs portal for.

jillianwilson:
Good point @florisvdg , it will be useful to have only one source of truth.

PR #49, opened 2023-07-04
Improve the shell script

edif2008:
Key changes in this PR:
- Use the new syntax for setting a step's output since [the old one is deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).
- Automatically fetch the latest stable version of the CLI.
- Add additional architectures for Linux
- Make the action fail if the GitHub Action runner's operating system is not supported.

Resolves: #27

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -39,6 +39,7 @@ unset_prev_secrets() {
 
 # Install op-cli
 install_op_cli() {
+  # Create a temporary directory where the CLI is installed
   OP_INSTALL_DIR="$(mktemp -d)"
   if [[ ! -d "$OP_INSTALL_DIR" ]]; then
     echo "Install dir $OP_INSTALL_DIR not found"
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -97,25 +104,27 @@ populating_secret() {
...
+  if [ "$INPUT_EXPORT_ENV" == "true" ]; then
     {
       # Populate env var, using heredoc syntax with generated identifier
-      echo "$env_var<<${random_heredoc_identifier}"
+      echo "$env_var<<${delimiter}"
       echo "$secret_value"
-      echo "${random_heredoc_identifier}"
+      echo "${delimiter}"
     } >> $GITHUB_ENV
     echo "GITHUB_ENV: $(cat $GITHUB_ENV)"
```
dustin-ruetz, line 120: Remove this blank line to make it clearer that the `if` and `else` clauses are related.

entrypoint.sh
```diff
@@ -97,25 +111,27 @@ populating_secret() {
...
 
-  if [ "$INPUT_EXPORT_ENV" == "true" ]; then
-    # To support multiline secrets, we'll use the heredoc syntax to populate the environment variables.
-    # As the heredoc identifier, we'll use a randomly generated 64-character string,
-    # so that collisions are practically impossible.
-    random_heredoc_identifier=$(openssl rand -hex 32)
+  # To support multiline secrets, we'll use the heredoc syntax to populate the environment variables.
+  # As the heredoc identifier, we'll use a randomly generated 64-character string,
+  # so that collisions are practically impossible.
+  delimiter="$(openssl rand -hex 32)"
```
dustin-ruetz, line 117: Appreciate the comments and the explanation on the call 🙏

Do you think it'd be worth linking to the [Multiline string](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings) section of the GitHub Docs here?

entrypoint.sh
```diff
@@ -39,30 +39,44 @@ unset_prev_secrets() {
...
+    elif [[ "$ARCH" == "x86_64" ]]; then
+      ARCH="amd64"
+    elif [[ "$ARCH" == "aarch64" ]]; then
+      ARCH="arm64"
+    fi
+
+    if [[ "$ARCH" != "386" ]] && [[ "$ARCH" == "amd64" ]] && [[ "$ARCH" == "arm" ]] && [[ "$ARCH" == "arm64" ]]; then
+      echo "Unsupported architecture for the 1Password CLI: $ARCH."
       exit 1
     fi
```
dustin-ruetz, line 68: Can we simplify this check by getting rid of this `if` chain and making the `echo` error message/`exit 1` the last condition in the above `if` block?

edif2008: There is a mistake in the conditional. Nice catch!
Fixed it with [f195e2d](https://github.com/1Password/load-secrets-action/pull/49/commits/f195e2df03423d2e879db28f05869acaf88aa6b3).

entrypoint.sh
edif2008: Added the link with [f195e2d](https://github.com/1Password/load-secrets-action/pull/49/commits/f195e2df03423d2e879db28f05869acaf88aa6b3).

dustin-ruetz (approved):

Issue #62, opened 2024-02-12
Fix YAML syntax error in README example


libutcher:
The code example in the README is currently broken because of the syntax used to print the string `Secret:` ahead of the output. YAML requires strings that include special characters like : to be [enclosed with quotes to print correctly as a string](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml).

PR #65, opened 2024-02-19
Fix example snippet in README

edif2008:
The code example in the README is currently broken because of the syntax used to print the string Secret: ahead of the output. YAML requires strings that include special characters like : to be [enclosed with quotes to print correctly as a string](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml). This MR fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ jobs:
           SECRET: op://app-cicd/hello-world/secret
 
       - name: Print masked secret
-        run: echo "Secret: $SECRET"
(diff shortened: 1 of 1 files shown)


dustin-ruetz (approved):
- It sounds like this PR will close [issue 62](https://github.com/1Password/load-secrets-action/issues/62) - can you update the PR description to reflect this?
- ✅ PR approved

PR #68, opened 2024-04-02
CONTRIBUTING.md

ingridcrant:
Adding a CONTRIBUTING.md. Here's an initial visual of it:
Screenshot 2024-04-02 at 1 58 56 PM

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,50 @@
+# Contributing
+
+Thank you for your interest in contributing to the 1Password load-secrets-action project 👋! Before you start, please take a moment to read through this guide to understand our contribution process.
+
+## Testing
+
+After following the steps below for signing commits, you can test against your PR with these steps:
+
+1. Create or use an existing repo to run the `load-secrets` GitHub Action
+2. In a workflow yaml file that uses the GitHub Action, modify the `uses: 1Password/load-secrets-action` line to be
+
+   ```
+   uses: 1Password/load-secrets-action@<branch-name>
+
+   # OR
+
+   uses: 1Password/load-secrets-action@<commit-hash>
+   ```
+
+3. Trigger the action, which now includes your changes
+
+## Documentation Updates
+
+If applicable, update the [README.md](./README.md) to reflect any changes introduced by the new code.
+
+## Sign your commits
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
+# Contributing
+
+Thank you for your interest in contributing to the 1Password load-secrets-action project 👋! Before you start, please take a moment to read through this guide to understand our contribution process.
+
+## Testing
```
edif2008, line 5: Since this action is a Node type action (i.e. executed using Node), we also have Typescript tests that can be run using `npm run test`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+1. Create or use an existing repo to run the `load-secrets` GitHub Action
+2. In a workflow yaml file that uses the GitHub Action, modify the `uses: 1Password/load-secrets-action` line to be
+
+   ```
+   uses: 1Password/load-secrets-action@<branch-name>
+
+   # OR
+
+   uses: 1Password/load-secrets-action@<commit-hash>
+   ```
```
edif2008, line 18: `[nitpick]`

Would it be better to separate these into two code blocks that can be copied?

```yaml
uses: 1Password/load-secrets-action@<branch-name>
```

or

```yaml
uses: 1Password/load-secrets-action@<commit-hash>
```

Also recommend using the `yaml` tag to the code snippet so that the syntax is properly highlighted: 
```
   ```yaml
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
+# Contributing
+
+Thank you for your interest in contributing to the 1Password load-secrets-action project 👋! Before you start, please take a moment to read through this guide to understand our contribution process.
+
+## Testing
+
+After following the steps below for signing commits, you can test against your PR with these steps:
+
+1. Create or use an existing repo to run the `load-secrets` GitHub Action
```
edif2008, line 9: `[nitpick]`

From what I know steps that contain sentences (like this one) should end with punctuation:
```suggestion
1. Create or use an existing repo to run the `load-secrets` GitHub Action.
```

Same applies to the other steps.

PR #81, opened 2024-11-03
Switch to /usr/bin/env bash in entrypoint.sh

Srylax:
Using hardcoded paths (`/bin/bash`) is generally a bad practice.
Please use the POSIX compliant `/usr/bin/env bash` instead.
Self-hosted runners that use a system like NixOS will currently fail to use the 1Password Action.


gskierk:
What is the status of this pull request? If it is already rejected, then please close it. It is confusing :(.

volodymyrZotov:
Closing as `configure` action now is written in JS

PR #87, opened 2024-12-16
Make workflow targets more specific

edif2008:
This ensures that:
- `Acceptance tests` workflow is run on pull requests (including the ones from external contributors)
- `Lint` is run on `main` branch.
- Workflows are executed only once on pull requests

**Context:** It was noticed that the `Acceptance tests` workflow was not executed on #82 which made the review cycle take way longer and required extra effort to identify bugs and regressions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,7 @@
-on: pull_request
+on:
+  push:
+    branch: [main]
+  pull_request:
 name: Lint
 
 jobs:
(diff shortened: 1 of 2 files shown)


bxb100:
> It seems that the test pipeline didn't execute on your branch. I have https://github.com/1Password/load-secrets-action/pull/87 open which should address this.

I'm sorry to misguide you, these actions are running on my repo, `test.yml` doesn't run because I disabled it.

---

image

I can't pass the test workflow if this repo uses `Settings -> Secrets and variables` to set secret values.

I believe https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment is the right way

Marton6 (approved):

PR #97, opened 2025-02-20
Run acceptance tests on fork

edif2008:
Currently an external contributor can't have the acceptance tests run on their PR because `pull_request` doesn't give access to the secrets needed for them.

Therefore, in this PR we create a new workflow that is identical to the one for existing acceptance tests, with the following differences:
- This workflow can be triggered with the command `/ok-to-test sha="<contributor's latest commit sha>"` by one of this repo's maintainers.
- After the acceptance tests finish, their result will be updated to the PR's list of checks.

### DIFF
--- a/.github/workflows/test-fork.yml
+++ b/.github/workflows/test-fork.yml
@@ -1,0 +1,92 @@
+on:
+  repository_dispatch:
+    types: [ok-to-test-command]
+name: Run acceptance tests [fork]
+
+jobs:
+  test-with-output-secrets:
+    if: |
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.args.named.sha != '' &&
+      contains(
+        github.event.client_payload.pull_request.head.sha,
+        github.event.client_payload.slash_command.args.named.sha
+      )
+    uses: ./.github/workflows/acceptance-test.yml
+    secrets: inherit
+    with:
+      secret: op://acceptance-tests/test-secret/password
+      secret-in-section: op://acceptance-tests/test-secret/test-section/password
+      multiline-secret: op://acceptance-tests/multiline-secret/notesPlain
+      export-env: false
+  test-with-export-env:
+    if: |
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.args.named.sha != '' &&
+      contains(
+        github.event.client_payload.pull_request.head.sha,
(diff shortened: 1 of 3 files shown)


.github/workflows/test-fork.yml
```diff
@@ -0,0 +1,206 @@
+on:
+  repository_dispatch:
+    types: [ok-to-test-command]
+name: Run acceptance tests [fork]
```
florisvdg, line 4: Is it not possible to have a single workflow file that listens for multiple `on` statements? One for our commits and one for forks?

edif2008: It is, but I've chosen separating the two for the following reasons:
- **Keeping the differences separate.** There are two key difference between the jobs in this workflow (for forks) and the ones for our commits:
  - Checkout: For our commits we only checkout the latest version of the branch. For forks, we checkout at a specific commit sha provided by us.
  - Conditions: For our commits the jobs are executed as usual. For forks, those will be skipped and later have their checks updated by a manual workflow (defined in this file) that is triggered by us.
- **Increased complexity.** When merging the two files into one, we'd need to add conditions to our steps and jobs, which can make things more complex and prone to edge cases.
- **Keeping things clean.** This workflow contains an extra step that is responsible to update the checks. It would be a bit weird to see this skipped every single time we make PRs. Also, knowing what will be executed manually versus in general is a cleaner distinction in separate files.

Now I can see why having two separate files can become a hassle from a maintenance standpoint, but I would take a step further and say that this can be simplified once we simplify each acceptance test job, since all 3 are over 80% identical. However, I would consider doing this outside of this PR.

AndyTitu:
Functionally tested this with @edif2008 on [another test Fork](https://github.com/AndyTitu/load-secrets-action/pull/3) and it works as expected.

AndyTitu:
It might also be worth deduplicating the GitHub action script we run in both test and test-fork.yml using this guide: https://joel-azemar.medium.com/github-actions-and-reusable-workflows-544f43e1b450

It would make it simpler to check the actual conditions introduced in the PR.

.github/workflows/test-fork.yml
edif2008: After refactoring the acceptance test with #99, this PR looks much cleaner. The similarities are even more visible now, but the reasoning behind keeping them separate still remains.

AndyTitu (approved):
Changes makes sense to me, thanks for adding @edif2008 !

PR #99, opened 2025-03-03
Refactor acceptance tests

edif2008:
This PR refactors the 3 acceptance tests to use the same reusable workflow. The difference between the 3 tests are the following:
- Secret references used
- Whether the secrets are passed as a step's output or as environment variables (for checking the values)

The reusable workflow is fairly identical to the one used in the `.github/workflows/acceptance-test.yml`. The differences are the following:
- The secrets used in the load-secrets-action come from the input passed to the reusable workflow.
- The `export-env` variable comes from the input passed to the reusable workflow.
- The steps for checking the loaded secrets are conditionally executed based on the `export-env` input passed to the reusable workflow.


.github/workflows/acceptance-test.yml
```diff
@@ -0,0 +1,76 @@
...
+        type: string
+      secret-in-section:
+        required: true
+        type: string
+      multiline-secret:
+        required: true
+        type: string
+      export-env:
+        required: true
+        type: boolean
```
edif2008, line 17: `[note for reviewers]`
With this refactor, I'm considering that it would be a better idea to no longer pass this as part of the test, but rather have this reusable workflow do both i.e. run the tests for providing secrets as a step's output and as env vars. These would be run in parallel in a the defined matrix.

With that, we would increase our acceptance tests from 9 to 12: a set of 6 for references with names, and another 6 for references with IDs.

AndyTitu: Personally I like this PR as it already is. We could consider doing what you proposed as part of a separate PR, but given that a contributor relies on another PR which relies on this one, I'd say that we don't increase the scope of this refactor.

AndyTitu (commented):
The refactor looks good to me. The pipeline still passes so no regression seem to have occurred.

AndyTitu (approved):

PR #100, opened 2025-03-05
Fix fork workflow

edif2008:
In #97 it was missed to adjust the reusable workflow to pull changes from the forked commit. Instead, now we pull from base repository, which doesn't contain the external contributor's changes.

I've also improved the way we reference the reusable workflow to ensure we're using a trusted reusable workflow that won't change often.

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -29,6 +29,26 @@ jobs:
     steps:
       - name: Base checkout
         uses: actions/checkout@v4
+        if: |
+          github.event_name != 'repository_dispatch' &&
+          (
+            github.ref == 'refs/heads/main' || 
+            (
+              github.event_name == 'pull_request' && 
+              github.event.pull_request.head.repo.full_name == github.repository
+            )
+          )
+      - name: Fork based /ok-to-test checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.client_payload.pull_request.head.sha }}
+        if: |
+          github.event_name == 'repository_dispatch' &&
+          github.event.client_payload.slash_command.args.named.sha != '' &&
+          contains(
+            github.event.client_payload.pull_request.head.sha,
+            github.event.client_payload.slash_command.args.named.sha
+          )
       - name: Launch 1Password Connect instance
         if: ${{ matrix.auth == 'connect' }}
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/acceptance-test.yml
```diff
@@ -29,6 +29,26 @@ jobs:
...
         uses: actions/checkout@v4
+        if: |
+          github.event_name != 'repository_dispatch' &&
+          (
+            github.ref == 'refs/heads/main' || 
+            (
+              github.event_name == 'pull_request' && 
+              github.event.pull_request.head.repo.full_name == github.repository
+            )
+          )
```
AndyTitu, line 40: What is the reason behind avoiding this checkout when both` github.event_name == 'repository_dispatch' ` and `github.ref == 'refs/heads/main'` happen?

.github/workflows/test-fork.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         github.event.client_payload.pull_request.head.sha,
         github.event.client_payload.slash_command.args.named.sha
       )
-    uses: ./.github/workflows/acceptance-test.yml
+    uses: 1password/load-secrets-action/.github/workflows/acceptance-test.yml@main
```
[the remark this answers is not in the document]
edif2008: According to the [documentation for calling reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#calling-a-reusable-workflow):
> You reference reusable workflow files using one of the following syntaxes:
>
> - `{owner}/{repo}/.github/workflows/{filename}@{ref}` for reusable workflows in public and private repositories.
> - `./.github/workflows/{filename}` for reusable workflows in the same repository.
>
> In the first option, `{ref}` can be a SHA, a release tag, or a branch name.

I don't think we have a default for this, thus the `main` branch name should stay in it.

.github/workflows/acceptance-test.yml
edif2008: Do we ever have the intention of running manual workflows on our branches? 🤔 
From my understanding we only do this on forked PRs (where we trigger manual workflows), therefore we shouldn't use this if we trigger workflows in other scenarios.

AndyTitu: Yeah, maybe not, the only case I had in mind is when running a GitHub action directly from the Actions panel. We don't have that yet but maybe we'll include it. But even that I think it just uses the workflow_dispacth event and not the repo_dispatch one.

AndyTitu (approved):

PR #109, opened 2025-08-05
Use op cli installer to enable Windows support

### LINKED ISSUE #46: load-secrets-action doesn't work on windows
```shell
Run 1password/load-secrets-action@v1
"C:\Program Files\Git\bin\sh.exe" -c D:\a\_actions\1password\load-secrets-action\v1/entrypoint.sh
/usr/bin/bash: line 1: D:a_actions1passwordload-secrets-actionv1/entrypoint.sh: No such file or directory
Error: The process 'C:\Program Files\Git\bin\sh.exe' failed with exit code 127
```

See action at: https://github.com/apache/incubator-opendal/actions/runs/5367357180/jobs/9737444126?pr=2525

volodymyrZotov:
This PR started to use [op-cli-installer](https://github.com/1Password/op-cli-installer/pull/1) package as dependency, to install CLI. This enables Windows support for `load-secrets-action`.

### Main changes
1. Use `installCliOnGithubActionRunner` function from  [op-cli-installer](https://github.com/1Password/op-cli-installer/pull/1) instead of `install-cli.sh` script.
2. `configure` action is refactored and now is written in pure JS, which helps wot enable Windows support.
3. Integration tests in github action piepeline are refactored to run on all runners and using different CLI versions (latest stable, latest beta, specific stable/beta, etc). Also test checks that `confiture` action set env vars properly and verifies installed CLI version.

Resolves #46


.github/workflows/acceptance-test.yml
edif2008, line 1: Moving matrices from this workflow to the parent one now makes the pipeline to have 51 checks (in parallel). That looks overwhelming for such a small action and hard to follow if something goes wrong. We should balance this a bit better in my opinion.

.gitignore
edif2008, line 1: [nit]

Missing newlline at end of file.

configure/index.js
edif2008, line 1: Why is this in JS and not in TS? 🤔

tests/assert-cli-version.sh
edif2008, line 1: Why is this test needed as part of this Action? It looks like something that is better suited for `install-cli-action`. 🤔

tests/assert-cli-version.sh
```diff
@@ -0,0 +1,30 @@
...
+    if [ "$1" == "non_beta" ]; then
+        conditional_path="!/beta/"
+    fi
+    # This long command parses the HTML page at "CLI_URL" and finds the latest CLI version
+    # based on the release channel we're looking for (stable or beta).
+    #
+    # The ideal call (i.e. 'curl https://app-updates.agilebits.com/check/1/0/CLI2/en/2.0.0/Y -s | jq -r .version')
+    # doesn't retrieve the latest CLI version on a channel basis.
+    # If the latest release is stable and we want the latest beta, this command will return the stable still.
+    OP_CLI_VERSION="$(curl -s $CLI_URL | awk -v RS='<h3>|</h3>' 'NR % 2 == 0 {gsub(/[[:blank:]]+/, ""); gsub(/<span[^>]*>|<\/span>|[\r\n]+/, ""); gsub(/&nbsp;.*$/, ""); if (!'"$1"' && '"$conditional_path"'){print; '"$1"'=1;}}')"
```
edif2008, line 18: I've noticed in `install-cli-action` that we fetch the latest version in a simpler manner. If we were to keep this here, I'd expect us to use a simpler version as well.

.github/workflows/test.yml
```diff
@@ -14,44 +14,86 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        version: [latest, latest-beta, 2.30.0, 2.30.0-beta.03]
+        auth: [connect, service-account]
+        exclude:
+          - os: macos-latest
+            auth: connect
+          - os: windows-latest
+            auth: connect
```
edif2008, line 64: `[nit]`

Might be complicated to achieve that in GitHub workflows, but can we extract this somehow to reuse it everywhere?

.github/workflows/acceptance-test.yml
edif2008: Just realized that we'd get the same number of checks regardless of where we define the matrices. The benefit of this approach is that external contributors will see the same number of checks as the maintainers, which is a nice touch of consistency.

tests/assert-cli-version.sh
volodymyrZotov: after discussion about it we decided to remove that script from here and rely on testing version in dependency package

.github/workflows/test.yml
volodymyrZotov: Good idea, thank you for pointing it out. Let's try to do this in a separate PR but for now keep it as it is.

edif2008 (approved):

package.json
```diff
@@ -9,6 +9,7 @@
 	},
 	"scripts": {
 		"build": "ncc build ./src/index.ts",
+		"build:configure": "ncc build ./configure/index.js -o ./configure/dist",
```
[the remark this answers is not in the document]
volodymyrZotov: introduced `build:all` command for that

PR #114, opened 2025-08-13
Set `export-env` input to `false` by default

### LINKED ISSUE #79: export-env security risk
Currently, `export-env` has a default value of `true` and is also suggested to be set as `true` in README.  

While it is convenient to do so, it makes the secrets available as ENVs to all the later steps, including the third-party GitHub Actions. This can easily lead to leaking secrets when using malicious or vulnerable GitHub actions.  
Thus, I think it should at least be mentioned in README.

Additionally, the usage of the step outputs of `load-secrets-action` should also be documented.

volodymyrZotov:
As we anyway plan to release [v3](https://github.com/1Password/load-secrets-action/pull/110), I think this is a good opportunity to introduce this breaking change.

So, in `v3` of the action the `export-env` input will defaults to `false`.
This will help to mitigate the next security concern. As exporting secret as env variable in the job will make it available for all future steps including 3d party actions. If some malicious 3d party action is used it can get an access to the secret from env var.

Though, the risk is low, the safer way for the action would be to defaults to `false` and let the users explicitly set `export-env` input, if they want to export secrets as env variables.

This PR also updates readme to show reccomended example of using secrets as step outputs as well as exporting secrets as env vars.

Resolves #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
 
 ## ✨ Quickstart
 
-### Use secrets from step output (recommended)
+### Export secrets as a step's output (recommended)
 
 ```yml
 on: push
@@ -56,10 +56,10 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Load secret
-        id: load_secret
         uses: 1password/load-secrets-action@v3
         with:
-          export-env: "true"
+          # Export loaded secrets as environment variables
+          export-env: true
         env:
           OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
           SECRET: op://app-cicd/hello-world/secret
--- a/README.md
+++ b/README.md
@@ -23,19 +23,41 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
 
 ## ✨ Quickstart
 
+### Use secrets from step output (recommended)
 ```yml
 on: push
 jobs:
   hello-world:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Load secret
-        uses: 1password/load-secrets-action@v2
+        id: load_secret
+        uses: 1password/load-secrets-action@v3
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,19 +23,43 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Load secret
-        uses: 1password/load-secrets-action@v2
+        id: load_secret
+        uses: 1password/load-secrets-action@v3
         with:
-          # Export loaded secrets as environment variables
```
edif2008, line 37: `[nit]`
I'd still keep the comment in the snippet for clarity.

README.md
```diff
@@ -23,19 +23,43 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
...
-          export-env: true
+          export-env: "true"
```
edif2008, line 62: `[non-blocking question]`
Is there any real difference between the two versions (before and after the change)? Or is this more around value safety?

README.md
```diff
@@ -23,19 +23,43 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
 
 ## ✨ Quickstart
 
+### Use secrets from step output (recommended)
```
edif2008, line 26: `[nit]`
We want to be consistent with the headers for the two cases. What we do here is export them in one way or another to be used for following steps.

```suggestion
### Export secrets as a step's output (recommended)
```

If you have another preference in terms of how the header should look like, make sure it's consistent for both cases.

README.md
```diff
@@ -23,19 +23,43 @@ Read more on the [1Password Developer Portal](https://developer.1password.com/do
...
 jobs:
   hello-world:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Load secret
-        uses: 1password/load-secrets-action@v2
+        id: load_secret
```
edif2008, line 59: `[nit]`
While it's good to have it, we don't need it for this use case, right? 🤔

edif2008 (approved):
Thank you for making this change. This will default to the safer way of providing secrets for following steps. 

Approved with a couple of nits.

README.md
volodymyrZotov: yes, we do not need it here, removing

README.md
volodymyrZotov: Right, we should use `true` as bool not as string. I simply revert it to be bool.

PR #115, opened 2025-08-14
Rename `version` input to `cli-version`

volodymyrZotov:
This PR renames `version` input to `cli-version` to make it more obvious about which version we are talking about.

It defaults to `latest` if not provided.

In the future releases when we will migrate from CLI to SDK, the `cli-version` will be deprecated. And we will introduce a warning about future depreciation, close to that moment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         github.event_name == 'pull_request' && 
         github.event.pull_request.head.repo.full_name == github.repository
       )
-    uses: 1password/load-secrets-action/.github/workflows/acceptance-test.yml@main
+    uses: 1password/load-secrets-action/.github/workflows/acceptance-test.yml@vzt/rename-version-input # TODO: temporary point to this branch so tests can run with update input
     secrets: inherit
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance-test.yml
```diff
@@ -15,7 +15,7 @@ on:
       export-env:
         required: true
         type: boolean
-      version:
+      cli-version:
```
MOmarMiraj, line 18: hmm if we are already recgonizing that the SDK is going to be supported here soon enough, is this change even necessary? Seems like more work as when the SDKs get introduced this has to be fixed again

volodymyrZotov: Great question!

I thought introducing `version` would be a nice to have feature, especially as we give it for free as of using [op-cli-installer package](https://github.com/1Password/op-cli-installer/blob/main/src/github-action/index.ts#L13). 

Having `version` input in the action would be useful in case `latest` CLI release has a regression, so users can start using older version of the CLI without the issue.

On another hand, as you fairly mentioned, as we move toward using SDK instead of CLI, that seems as additional work. Considering the fact that current `load-secrets-action@v2` uses `latest` cli version as default, maybe we do not need to introduce this additional input at all and should remove it?

PR #154, opened 2026-03-24
fix 1Password cli installer error on Windows

superteppo:
Override install() in WindowsInstaller to rename the downloaded file with a .zip extension before extraction.

On Windows, @actions/tool-cache’s downloadTool() saves downloads using a random UUID filename with no extension. When pwsh is unavailable and only legacy powershell is present, extractZip() falls back to Expand-Archive, which requires the file to have a .zip extension. Before this change, extraction failed because the downloaded file had no .zip suffix.


JillRegan (approved):
Thanks @superteppo, tested and confirmed on Windows. LGTM! 👍

Issue #101, opened 2025-03-06
Performance optimization


zshannon:
Hi there, is there something I could do differently to speed up reading secrets? It takes 14s to read 14 secrets.

gskierk:
It's already reported here https://github.com/1Password/load-secrets-action/issues/91.

I ended up having my own shell file to read secrets concurrently - it reads around 30 secrets in 3 seconds.

SimonBarendse:
I'm closing this as duplicate of #91 with work in progress to address in #82

gskierk:
@SimonBarendse This PR seems to not solve this kind of problem. https://github.com/1Password/load-secrets-action/pull/82#discussion_r1899046609

PR #163, opened 2026-04-24
[WIP] Parallelize env var reads

### LINKED ISSUE #91: Reading refs concurrently
Reading 29 refs (using "op read" one by one) now takes 20-25 seconds. It would be great if we could cut this time by reading all refs concurrently (Promise.all?). Reading them this way takes around 2-3 seconds. Is it possible to implement in this action?

jhnns:
Adds support for loading secrets from an environment file in batches, improving the performance for workflows that rely on multiple secrets.

Closes #91 #101

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -35472,6 +35472,9 @@ const installCliOnGithubActionRunner = async (version) => {
 
 
 
+;// CONCATENATED MODULE: external "node:fs"
+const external_node_fs_namespaceObject = require("node:fs");
+var external_node_fs_default = /*#__PURE__*/__nccwpck_require__.n(external_node_fs_namespaceObject);
 ;// CONCATENATED MODULE: ./package.json
 const package_namespaceObject = {"rE":"4.0.0"};
 ;// CONCATENATED MODULE: ./src/constants.ts
@@ -35488,6 +35491,9 @@ const authErr = `Authentication error with environment variables: you must set e
 
 
 
+
+
+const envFileKeysEnvVar = "OP_KEYS_JSON";
 const validateAuth = () => {
     const isConnect = process.env[envConnectHost] && process.env[envConnectToken];
     const isServiceAccount = process.env[envServiceAccountToken];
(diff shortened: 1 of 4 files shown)


jhnns:
Hi everyone 👋,

first of all: thank you for this GitHub action! I really like the approach of storing my .env secrets in 1Password :)

This PR is a PoC implementation for a performance improvement:

In our application, we have about 15 environment variables that need to be loaded from 1Password. In the current implementation, this takes about 25–30 seconds. It appears that "Populating variables" is executed sequentially. For us, this is kind of unacceptable because the whole action just takes 1 minute (30 seconds of it just loading env variables).

I know that this has been brought up before (#91 #101), but the referenced PR #82 has been closed. I also think that this could be fixed now without waiting for a complete rewrite.

This PR is a proof-of-concept implementation of how parallel loading of environment variables might work. As a disclaimer upfront: this implementation was generated by AI. It includes some changes that I believe are useful but have nothing to do with this fix.

I just wanted to know if you're interested in this change and if you think the approach is generally okay before I invest any more time in it.

I tested this PR in our setup and it brought the action down to 4s.

dist/index.js
jhnns, line 1: Do you want the `dist` to be included in the PR?

src/index.ts
```diff
@@ -39,7 +48,7 @@ const loadSecretsAction = async () => {
 		if (error instanceof Error) {
 			message = error.message;
 		} else {
-			String(error);
+			message = String(error);
```
jhnns, line 51: Unrelated, but useful change. I can open a separate PR for it.

src/utils.test.ts
```diff
@@ -141,11 +150,13 @@ describe("loadSecrets", () => {
...
-			name: "1Password GitHub Action",
-			id: "GHA",
-		});
-		expect(exec.getExecOutput).toHaveBeenCalledWith('sh -c "op env ls"');
+		expect(setClientInfo).toHaveBeenCalledWith(
+			expect.objectContaining({
+				name: "1Password GitHub Action",
+				id: "GHA",
+			}),
+		);
```
jhnns, line 158: Unrelated change. I can revert that :)

src/utils.test.ts
```diff
@@ -141,11 +150,13 @@ describe("loadSecrets", () => {
...
+		expect(exec.getExecOutput).toHaveBeenCalledWith("op", ["env", "ls"]);
```
jhnns, line 159: Also unrelated, see my comment at the implementation

src/utils.ts
```diff
@@ -57,6 +61,84 @@ export const extractSecret = (
...
+			nodeScript,
+		],
+		{
+			silent: true,
+			env: {
+				...process.env,
+				[envFileKeysEnvVar]: JSON.stringify(envNames),
+			},
+		},
+	);
```
jhnns, line 124: I can put `nodeScript` into a dedicated file. Do you think the approach of "abusing" `op run` is ok?

src/utils.ts
```diff
@@ -68,7 +150,7 @@ export const loadSecrets = async (shouldExportEnv: boolean): Promise<void> => {
 	// Load secrets from environment variables using 1Password CLI.
 	// Iterate over them to find 1Password references, extract the secret values,
 	// and make them available in the next steps either as step outputs or as environment variables.
-	const res = await exec.getExecOutput(`sh -c "op env ls"`);
+	const res = await exec.getExecOutput("op", ["env", "ls"]);
```
jhnns, line 165: This change is also unrelated.

Personally, I think the change makes sense, but maybe there's a specific reason for `sh -c`? I can do a separate PR if you're interested.

PR #164, opened 2026-05-20
Verify 1Password CLI signature on install

JillRegan:
## Summary

Adds signature verification for the 1Password CLI binary the action downloads, addressing concern that the action did not validate the authenticity of artifacts downloaded from `cache.agilebits.com`.

Each platform uses the verification mechanism from 1Password's docs (https://www.1password.dev/cli/verify). 

## How each platform verifies

- **macOS**: `pkgutil --check-signature` - pinned to team ID `2BUA8C4S2C` and a SHA-256 allowlist of accepted leaf certs.
- **Linux**: `gpg --verify` against a bundled public key and pinned to fingerprint `3FEF9748469ADBE15DA7CA80AC2D62742012EA22`.
- **Windows**: `Get-AuthenticodeSignature` - requires `Status=Valid` and a Subject containing `CN=Agilebits`.

## Breaking changes (this should be semver-major update)

1. **Verification is now mandatory.** Before this change the action installed whatever the CDN served. After this change a binary that fails verification is rejected.
2. **Linux runners must have `gpg` on PATH.** GitHub-hosted Ubuntu runners include it by default. Custom or minimal container images may need updating.

## Why the Linux key is bundled

The Linux public key lives at `src/op-cli-installer/github-action/cli-installer/linux-signing-key.asc` (ncc auto-bundles it into `dist/` next to `index.js`). This eliminates a runtime dependency on `keyserver.ubuntu.com` or `downloads.1password.com`.

## Failure path verification

To confirm verification actually runs against real binaries in CI I created a throwaway branch and replaced a single character in each pinned trust anchor to view verification failure.
- [linux](https://github.com/1Password/load-secrets-action/actions/runs/26249246636/job/77255719514)
- [macos](https://github.com/1Password/load-secrets-action/actions/runs/26249447457/job/77256416828)


rishiy15 (approved):

src/op-cli-installer/github-action/cli-installer/windows-signature.ts
```diff
@@ -0,0 +1,60 @@
...
+	const status = fieldValue("Status=");
+	if (status !== "Valid") {
+		throw new Error(
+			`Authenticode status is ${status ?? "unknown"}, expected Valid.\nGet-AuthenticodeSignature output:\n${output}`,
+		);
+	}
+
+	// Confirm the signer is AgileBits, not some other publisher.
+	const subject = fieldValue("Subject=") ?? "";
+	if (!subject.includes(`CN=${WINDOWS_SIGNER_SUBJECT_CN}`)) {
```
bertrmz: This could use a closing delimiter; the current form would match `CN=AgilebitsAttackerCorp` as well.

bertrmz (approved):

JillRegan:
I’m going to hold off on merging this for the moment while I dig into some flaky tests.

For folks who want CLI signature verification now, the [latest](https://github.com/1Password/install-cli-action/releases/tag/v4.0.0) `install-cli-action` already performs that verification. You can use it earlier in your workflow to install the CLI, and load-secrets-action will skip installing it if it’s already available. 

Will be following up with this next week to confirm and test further.

PR #169, opened 2026-06-16
Support Workload Identity auth via 1Password SDK (private beta)

JillRegan:
Adds a new way to authenticate the action with Workload Identity. Instead of storing a Service Account token or Connect credentials, the action exchanges the GitHub Actions OIDC token for short-lived 1Password access.

To use it, set three environment variables on the step:
- `OP_WORKLOAD_ID`
- `OP_ENVIRONMENT_ID`
- `OP_INTEGRATION_KEY`

When all three are set, the action loads every variable from the configured 1Password environment. If none are set, it falls back to the existing Service Account / Connect flow.

### Changes

- New auth path via the @1password/sdk package (`src/sdk-client.ts`).
- Validation for the new config: errors clearly if only some of the three vars are set, or if Workload Identity is mixed with another auth method.
- Unset env: a cleanup step (unset-previous) works without credentials, so Workload Identity users can remove secrets in a later step.
- Tests for the new config logic and SDK client.
- E2E job covering the Workload Identity flow.
- README section documenting the feature (marked private beta).

⚠️ This is a private beta release, hence why this PR targets a beta branch as it will be released from there.


src/sdk-client.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const loadSecretsFromSDK = async (
+	workloadId: string,
+	environmentId: string,
+	integrationKey: string,
+	shouldExportEnv: boolean,
+): Promise<void> => {
+	// Temporary fix: strip base64 padding from integrationKey — this will eventually be handled by the SDK core itself
+	const customerManagedSecret = integrationKey.replace(/=+$/, "");
```
bertrmz, line 18: To ensure Github [masks this value](https://github.com/actions/toolkit/blob/main/packages/core/README.md#setting-a-secret) if it ever appears in logs:

```suggestion
	const customerManagedSecret = integrationKey.replace(/=+$/, "");
	core.setSecret(customerManagedSecret);
```

bertrmz (approved):

PR #170, opened 2026-06-17
Release v5.0.0-beta.1

JillRegan:
Release v5.0.0-beta.1

### Feature

- Add Workload Identity authentication using the GitHub Actions OIDC token (private beta) ([#169](https://github.com/1Password/load-secrets-action/pull/169))

## Notes for reviewer:

Had to make two other minor changes in this PR:

- Stripped the prerelease suffix before `semverToInt` as the `5.0.0-beta.1` version bump produced an invalid CLI build number (400 errors).
- Updated the empty-string `setSecret` tests to assert "not called with an empty string" instead of "not called at all", since the integration key is now always masked.

### DIFF
--- a/src/sdk-client.test.ts
+++ b/src/sdk-client.test.ts
@@ -77,17 +77,6 @@ describe("loadSecretsFromSDK", () => {
 		);
 	});
 
-	it("masks the integration key and strips its base64 padding", async () => {
-		await loadSecretsFromSDK(
-			workloadId,
-			environmentId,
-			"integration-key==",
-			false,
-		);
-
-		expect(core.setSecret).toHaveBeenCalledWith("integration-key");
-	});
-
 	describe("when secret value is empty string", () => {
 		beforeEach(() => {
 			mockGetVariables.mockResolvedValue({
--- a/src/sdk-client.test.ts
+++ b/src/sdk-client.test.ts
@@ -78,7 +78,12 @@ describe("loadSecretsFromSDK", () => {
 	});
 
 	it("masks the integration key and strips its base64 padding", async () => {
-		await loadSecretsFromSDK(workloadId, environmentId, "integration-key==", false);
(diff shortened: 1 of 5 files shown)


src/sdk-client.test.ts
```diff
@@ -93,14 +104,14 @@ describe("loadSecretsFromSDK", () => {
 			);
 
 			expect(core.setOutput).toHaveBeenCalledWith("EMPTY_SECRET", "");
-			expect(core.setSecret).not.toHaveBeenCalled();
```
JillRegan, line 96: Had to update these as didn't see the failure for the [last commit](https://github.com/1Password/load-secrets-action/pull/169/changes/64ccbe04e832dfcb9e2cc767848de40a659548ab) on the other PR. Masking the integration key means setSecret now always runs, so the test just need to be changed from asserting it was never called to asserting it was never called with an empty string.

src/utils.ts
```diff
@@ -111,11 +111,14 @@ export const extractSecret = (
...
-	// Pass User-Agent Information to the 1Password CLI
+	// Pass User-Agent Information to the 1Password CLI.
+	// Strip any prerelease suffix ("-beta.1") since semverToInt only
+	// accepts a plain MAJOR.MINOR.PATCH; a prerelease produces an invalid build.
+	const [releaseVersion] = version.split("-");
 	setClientInfo({
 		name: "1Password GitHub Action",
 		id: "GHA",
-		build: semverToInt(version),
+		build: semverToInt(releaseVersion ?? version),
```
JillRegan, line 119: Bumping the version in package.json to a prerelease (5.0.0-beta.1) broke the build number, since semverToInt can't parse the -beta.1 suffix. So I needed to strip the prerelease suffix before converting.

scottisloud (approved):

PR #173, opened 2026-07-13
load-secrets-action: Add Docker Hub fallback for CLI version resolution

JillRegan:
## Summary

Resolving the latest CLI version currently depends on a single request to `app-updates.agilebits.com`. When that endpoint has an outage the action fails and breaks users' CI.

This adds two fallback tiers so version resolution survives an outage:

1. **app-updates** (unchanged primary)
2. **Docker Hub**  the `1password/op` image tags, on separate infrastructure
3. **Baked-in versions** a pinned constant refreshed at release time

If a tier fails, resolution falls through to the next. The action should not fail as final constant is pinned version.

## How

- `getLatestVersion` now tries `app-updates` first and falls back to the Docker Hub tags API.
- Docker Hub returns all tags, so the fallback filters to the requested channel (`N.N.N` for stable, `N.N.N-beta.N` for beta) and picks the highest version. Beta tags aren't valid semver (leading zero), so they're normalized only for comparison and returned unchanged.
- Baked-in tier: `src/op-cli-installer/version/fallback-versions.ts` holds the last known stable and beta versions, regenerated from app-updates via `npm run op-version` (run at release time, see release instructions).

## Testing

- Unit tests cover primary success.
- Two E2E jobs (matrixed over `latest` and `latest-beta`) blackhole hosts via `/etc/hosts` and verify the correct version still installs

In the E2E test [here](https://github.com/1Password/load-secrets-action/actions/runs/29342376200/job/87117206962?pr=173) you can see the warning that we fell back to Docker and successfully resolved the correct version.

```
Run ./
Authenticated with Service account.
Validating version number: 'latest-beta'
Version number 'latest-beta' is valid
Resolving version: latest-beta
Getting latest-beta version number
Warning: Could not resolve latest-beta version from app-updates.agilebits.com (TypeError: fetch failed); falling back to Docker Hub
Downloading 1Password CLI from: https://cache.agilebits.com/dist/1P/op2/pkg/v2.37.0-beta.01/op_linux_amd64_v2.37.0-beta.01.zip
Installing 1Password CLI
```

In the E2E test [here](https://github.com/1Password/load-secrets-action/actions/runs/29342376200/job/87117206868?pr=173) you can see the warning that we fell back to pinned version and successfully resolved the correct version.

```
Run ./
Authenticated with Service account.
Validating version number: 'latest'
Version number 'latest' is valid
Resolving version: latest
Getting latest version number
Warning: Could not resolve latest version from app-updates.agilebits.com (TypeError: fetch failed); falling back to Docker Hub
Warning: Could not resolve latest version from Docker Hub (TypeError: fetch failed); using the version baked in at build time
Downloading 1Password CLI from: https://cache.agilebits.com/dist/1P/op2/pkg/v2.35.0/op_linux_amd64_v2.35.0.zip
Installing 1Password CLI
/usr/bin/unzip -o -q /home/runner/work/_temp/1006306c-ee30-4fa0-aede-157834c4ef70
1Password CLI installed

```

### ⚠️ Note for Reviewer

I also refactored the E2E test setup. E2E tests for the action were redundant and flakey and often led to rate limiting issues on the test account (these tests had ~200 read requests each run). There was no need to be testing that many different combinations of the same read so i refactored to reduce it to around 30 reads. The following now runs: 

Service Account (ubuntu, latest, export-env=false)
Service Account (ubuntu, latest, export-env=true)
Service Account smoke (macOS, latest)
Service Account smoke (Windows, latest)
Connect (ubuntu, latest)
Install pinned CLI (ubuntu, 2.30.0)
Docker Hub fallback (ubuntu, latest)
Docker Hub fallback (ubuntu, latest-beta)
Baked-in fallback (ubuntu, latest)
Baked-in fallback (ubuntu, latest-beta)


.github/workflows/e2e-tests.yml
```diff
@@ -25,16 +25,17 @@ on:
         required: true
 
 jobs:
```
JillRegan, line 27: The changes in this E2E file are a refactor to reduce the amount of redundant testing we were doing. Essentially we were running tests on every OS, both export-env values, and for both pinned and latest CLI versions. This was roughly ~200 read requests and the tests were flakey and often hit rate limiting issues.

There is no need to do a full test for all those different combinations. Full resolution coverage (all secret shapes, both delivery paths, both export-env branches) now runs once on ubuntu, with a lightweight per OS smoke test to confirm the pipeline still works on macOS and Windows. The main reason for running on every OS was to confirm the CLI installs correctly on each platform, not to re-test secret resolution. So we keep per-OS install/smoke coverage but only do the full secret-loading suite once, on ubuntu.

We now run as follows:
- Service Account (ubuntu, latest, export-env=false) - full secret resolution test
- Service Account (ubuntu, latest, export-env=true) - full secret resolution test
- Service Account smoke (macOS, latest) 
- Service Account smoke (Windows, latest)
- Connect (ubuntu, latest) - full secret resolution test
- Install pinned CLI (ubuntu, 2.30.0)
- Docker Hub fallback (ubuntu, latest)
- Docker Hub fallback (ubuntu, latest-beta)
- Baked-in fallback (ubuntu, latest)
- Baked-in fallback (ubuntu, latest-beta)

This reduces reads to ~30 and is much more reliable.

src/op-cli-installer/version/helper.ts
```diff
@@ -1,21 +1,112 @@
...
+	const latestBeta = json.CLI2?.beta?.version;
 	const version =
 		channel === ReleaseChannel.latestBeta ? latestBeta : latestStable;
 
 	if (!version) {
-		core.error(`No ${channel} versions found`);
+		throw new Error(`No ${channel} versions found`);
+	}
+
+	validateVersion(version);
```
bertrmz, line 75: This validation can be applied to the Docker Hub results as well, & the baked-in fallbacks too just for completeness.

package.json
```diff
@@ -21,6 +21,7 @@
 		"test:coverage": "npm run test -- --coverage",
 		"test:watch": "npm run test -- --watch",
 		"typecheck": "tsc",
+		"op-version": "node scripts/update-fallback-versions.mjs",
```
bertrmz, line 24: To ensure this stays in sync, we could have a CI check when releasing that runs `npm run op-version`, and fails if `fallback-versions.ts` changed.

.github/workflows/e2e-tests.yml
```diff
@@ -221,30 +283,198 @@ jobs:
...
+
+      - name: Assert the fallback resolved the same version app-updates would have
+        run: |
+          expected="${{ steps.expected.outputs.version }}"
+          actual="$(op --version)"
+          if [ "$actual" != "$expected" ]; then
+            echo "Expected CLI version '$expected' but got '$actual'"
+            exit 1
+          fi
+          echo "Docker Hub fallback resolved $actual as expected"
```
bertrmz, line 405: Possible edge cases here could be that during a release window, or if a release succeeds on one channel and not the other, the 2 available versions might not coincide. We could avoid that failure case by asserting a weaker invariant, like just checking that the resolved version is non-empty, and that it's valid semver / matches the channel shape.

I wanted to raise the edge cases here, just to consider whether the CLI release process would make them plausible. But happy to leave this as-is, if it seems unlikely to bite us.

scottisloud (commented):
No feedback to add beyond what Bert's already flagged!

src/op-cli-installer/version/helper.ts
JillRegan: Good point, updated 👍

package.json
JillRegan: I have added a job that runs on creation of release branch to check this.

.github/workflows/e2e-tests.yml
JillRegan: Good call out. This could cause a failure if one release flow doesn't land on a certain platform but I think for now I'd like to leave the strict check and see how it behaves. 

If this starts failing often that also tells us how often the two sources actually diverge in practice which hopefully should't happen.

.github/workflows/check-fallback-versions.yml
```diff
@@ -0,0 +1,24 @@
...
+name: Check fallback versions are fresh
+
+# Runs when a release branch is created/updated to check pinned CLI version
+jobs:
+  check-fallback-versions:
+    name: "Check baked-in fallback versions are up to date"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v4
```
bertrmz: This can use `setup-node@v6` to align with the other workflows.

dist/index.js
bertrmz, line 1: With the latest code changes, on more `dist/` rebuild is needed.

JillRegan: Oops, thanks for catching!

bertrmz (approved):

PR #183, opened 2026-07-22
Prepare Release v4.1.1

JillRegan:
### Features
- Add fallback version resolution so if app-updates.agilebits.com is unavailable, the action now falls back to Docker Hub and then a baked-in pinned version. (#173 )

### Security
- Harden CI security: add `StepSecurity` harden runner and pin GitHub Actions to commit SHAs across workflows, restrict workflow permissions, and add Dependabot config. (#174 )


scottisloud (approved):
LGTM

